PR #5, opened 2020-09-03
Adds a new how to use page to the top nav bar

alannaemmrie:
Adds a how to use section on a separate page.  Information comes from usage on README and styling matches with original index.html.


shahednasser:
Hi there! Thank you for your help!

I think it would look best if "How to Use" was on the same page, before the list of buttons. 

Also, I think the font we're using does not look good for paragraph text (i.e something other than headings) so maybe for the content of the "How to Use" page we can use a different font from google font. If you can find something nice then please do and if not just put the section on the home page and I'll find a better font later on.

If you can also use something like [prism](https://prismjs.com/#basic-usage-cdn) to display the code better I think that would be nice, although not required. Just putting the content of the section on the home page is enough.

shahednasser (changes requested):
Put "How to Use" section in `index.html` instead of new page.

Optional: 
- find a better font for the text (not headings, headings should have the current font)
- Use Prism.js to format code blocks

alannaemmrie:
Ok! I will work on this right now. Do you have any suggestions for fonts you think would match the overall theme, or should I just look around and see if there's anything nice?

shahednasser:
@alannaemmrie I don't have anything in mind but I think a sans-serif style font would be best. Pick whatever you like most!

Issue #6, opened 2020-09-03
Make Download link smaller on small devices


shahednasser:
Make the download link in the navbar smaller on devices with max width 767px.
You can make the changes in `assets/css/index.css`

PR #9, opened 2020-09-04
Reduce download link to 1.15rem font size at xs screen size (<768px).

bjthompson805:
Resolves issue #6 to make download link smaller on devices up to 767px.

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -116,16 +116,23 @@ hr {
   text-decoration: underline;
 }
 
-.navbar .download-link {
-  margin-right: 1em;
+.navbar {
   font-size: 1.5rem;
 }
 
-.download-link-xs {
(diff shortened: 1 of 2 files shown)


shahednasser (changes requested):
Thank you for your work!

I like the approach you took. However, on smaller devices there will be like an extra line next to "Download" since the font size for the container `a` element is a larger font size. I think it's best if we use the `@media` queries approach.

Changes to make to `assets/css/index.css`:
- `@media screen and (min-width: 768px)` the font size for `.download-link` should be 1.5rem
- `@media screen and (max-width: 767px)` the font size for `.download-link` should be 1.15rem

This way, no changes need to be made to the original `index.html`

shahednasser:
Great! Thank you.

@all-contributors please add @bjthompson805 for code

Issue #8, opened 2020-09-04
Add tagline under logo


shahednasser:
Add a tagline under the logo that says "Made with ❤︎". Either use a sans-serif font (not the current font being used for the logo) or find something on Google Fonts.

Any style additions you make, please add them in `assets\css\index.css`. If you add a Google Font, add it at the end of the `<head>` tag.

valeriaoshiro:
Hello, I would like to work on this. Thank you

shahednasser:
Hi @valeriaoshiro , someone already is working on this issue. Please try other issues, I would love your help on them!

shahednasser:
Currently, no one is working on this issue if someone wants to work on it.

PR #10, opened 2020-09-04
Added Tagline Under Logo ( Issue: #8)

SeshaiHari:
Added a tagline under the logo saying 'Made with ❤︎'.

Issue: #8


shahednasser (changes requested):
Thank you for your work! I really liked what you did.

I just need you to do two things please:
1. Change the color of the heart from red to `#f74545`
2. there seems to be a problem in the `index.html` there's a spacing or some character after `<html>` tag so please remove it so that the `<head>` tag is right after the `<html>` tag.

SeshaiHari:
> Thank you for your work! I really liked what you did.
> 
> I just need you to do two things please:
> 
> 1. Change the color of the heart from red to `#f74545`
> 2. there seems to be a problem in the `index.html` there's a spacing or some character after `<html>` tag so please remove it so that the `<head>` tag is right after the `<html>` tag.

Very Sorry the Error in HTML file. Now the fix has been applied.

shahednasser:
It's ok but there still seems to be issues in the spacing in `index.html`. I'm not sure if it's because of an editor you're using.

Try to copy your work in `index.html` for the tagline, then `git checkout index.html` and try adding what you copied again. Maybe try using a different editor. If the problem persists just commit the CSS file and I will add the tagline to the HTML.

shahednasser:
Closing this PR request due to inactivity

Issue #17, opened 2020-09-05
Add section separator after How to Use section


shahednasser:
All button sections have a separator between them which is `<hr class="secondary-hr">`.

Please add it after the "How to Use" section before "Basic Buttons"

PR #18, opened 2020-09-05
Added separator after How To Use section

fluorspar20:
Addressed #17 

Issue:


shahednasser:
Hi there! Thank you for your help. However, this issue is meant for first-timers only. If you want you can help with other issues that don't have the "first-timers-only" label I would appreciate it!

PR #24, opened 2020-09-05
added seperator after "How to use section" and "Basic Buttons"

ItsmeSauravSonu:
added seperator after "How to use section" and "Basic Buttons"
issue #17 

Issue:


shahednasser:
Great! Thank you.

@all-contributors please add @ItsmeSauravSonu for code

PR #30, opened 2020-09-06
Sidebar scroll

isonnymichael:
I use some trick css by adding padding-top and margin-top, so when we click the sidebar, the content will fully covered

Issue: Navigating onclick sidebar bug


shahednasser:
Nice and thank you for reporting the bug.

@all-contributors please add @isonnymichael for bug

Issue #27, opened 2020-09-05
Add a "Scroll to top" button


shahednasser:
As the website can get lengthy, we need a button on the bottom right corner that appears after scrolling a little and that can take the user back to top on click.

Pratyush-Saxena:
@shahednasser can i work on this issue

shahednasser:
Yes go ahead!

Pratyush-Saxena:
@shahednasser pls check out my pr #34

PR #34, opened 2020-09-06
scroll-top btn added

Pratyush-Saxena:
Scroll to top button added , visit https://pratyush-saxena.github.io/sbuttons/ to see the changes

Issue: #27 resolved

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -246,7 +246,7 @@ hr {
 
 }
 .scroll-top i {
-  font-size: 45px;
+  font-size: 35px;
   display: block;
   color: #3A91FF;
   background: none;
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -220,9 +220,7 @@ hr {
   .navbar .download-link {
     font-size: 1.15rem;
   }
-  .scroll_to_top_btn{
-    font-size: 35px;
-  }
+  
   
 }
 
@@ -230,7 +228,6 @@ hr {
   .navbar .download-link {
     font-size: 1.5rem;
   }
-  
 }
 
 .social-btn {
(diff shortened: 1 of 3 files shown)


Pratyush-Saxena:
added sroll-to-top button

shahednasser (changes requested):
Great work! Just a few changes please:

1. Make the button's width/height 40px, the border-radius 50%, and the color of the icon white
2. Put the button on the right side of the page
3. When the page first opens you can see the button. If you can put the code for the `scroll` event in a function and call it on page load that would be great. You can see an example of what I mean with the `checkNavbar` and `checkActiveCategory` functions.

Pratyush-Saxena:
@shahednasser , I have made Changes as per your request pls check it out
https://pratyush-saxena.github.io/sbuttons/

shahednasser:
Great! There's something still pending which is that the `.scroll-top` element is still appearing on page load.

https://github.com/shahednasser/sbuttons/blob/1042d9cc7d0c52c620fffc38c91f39d8c794116a/assets/js/index.js#L66-L70

Move this part into a function under the `checkNavbar` function. Name the function `checkScrollTop` or anything you feel fitting. Then, call the function under the `checkNavbar` and `checkActiveCategory` functions on lines 57 and 61:

https://github.com/shahednasser/sbuttons/blob/1042d9cc7d0c52c620fffc38c91f39d8c794116a/assets/js/index.js#L56-L62

There's already a scroll event so let's not create an extra one.

Also, please change the `font-size` for `.scroll-top i` to 35px and set `right` for `.scroll-top` to 20px instead of setting `left` (remove the left property)

Pratyush-Saxena:
@shahednasser .....i have done all the changes stated above  this time  ..just one thing is there that I m unable to fix....is that.. As u asked me to set right property for .scroll-to instead of left.....so I tried to use float : right which isn't working....that why I used left margin to display the button in right side of window

shahednasser:
ok @Pratyush-Saxena no problem. Keep the `left` for now and commit the changes you made.

Pratyush-Saxena:
@shahednasser hope now its all done in the way you want....

shahednasser:
Great @Pratyush-Saxena thank you

@all-contributors please add @Pratyush-Saxena for code

Issue #38, opened 2020-09-06
Add a note about using font awesome for some buttons


shahednasser:
Currently, in our Usage docs and on the "How to Use" section of the website, we don't have a note about font awesome being needed for some buttons like Icon buttons and Social media login button. 

Add a little explanation of when font awesome is needed and include a code block of adding font awesome using their cdn. It should be added to both README and "How to Use" section.

lee00286:
Hello, may I work on editing README and How to Use section?

shahednasser:
yes, please!

shahednasser:
Resolved by #52

PR #52, opened 2020-09-08
Add a note about using font awesome

lee00286:
* Add a note about font awesome cdn in README.md
* Add a note section about font awesome cdn in index.html 

Issue: Add a note about using font awesome for some buttons #38


shahednasser (changes requested):
Great. Just a few remarks on the text: 
1. Instead of saying in case of using Download icon and Social icon, just say something like "In case of using buttons that have icons in them, make sure to include font awesome's CDN in the `<head>` tag".
2. Remove the text about logging in to font awesome. After righting something like the text above just put the code block for including font awesome.

Please follow these remarks for both the README and the HTML.

shahednasser:
Thank you!

@all-contributors please add @lee00286 for code and doc

PR #60, opened 2020-09-09
Added a basic pink button

yashgupta18:
Issue:


shahednasser:
I like it! Can you add examples for pink for the rest of the button types as well? like for hover, animated, etc...  
You just need to add `@darkPink` for hover, for the rest just follow the previous examples.  
Let me know if it's any trouble for you, we'll just merge it as it is now.

yashgupta18:
No Problem. Will Do it by tomorrow.

PR #62, opened 2020-09-09
removed default box-shadow

shubhampandeyhld:
After clicking the buttons an default box-shadow was added which was not looking good.

Issue:

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -118,6 +118,7 @@ section {
 }
 
 .button-container .sbtn {
+  box-shadow: none;
   margin-bottom: 1em;
   font-size: 1.5rem;
 }
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -122,6 +122,10 @@ section {
   margin-bottom: 1em;
   font-size: 1.5rem;
 }
+.button-container .sbtn :hover{
+  border:2px solid red !important;
+}
+
 
 .button-caption {
   color: #333;
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -121,9 +121,6 @@ section {
   margin-bottom: 1em;
   font-size: 1.5rem;
 }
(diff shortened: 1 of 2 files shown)


shahednasser:
Thank you for contributing! The box-shadow for the buttons is important on click to show that the button was clicked (this is more for user experience). If you think the current box-shadow doesn't look good, then please provide an alternative that can show the user that they indeed clicked the button.

shubhampandeyhld:
Okay I'll come with an alternative ,meanwhile I am also adding og tags for seo but their is no site logo present here.

shahednasser:
Yes, we currently don't have one. There's an issue about it (#2 ) but nobody has contributed yet, unfortunately.  
If you're working on og tags then just put the tags in for now and once there's a logo we'll add it in.

shahednasser:
@shubhampandeyhld  I saw you added the tags for SEO. Will you add something else regarding the box shadow? If not please remove the current changes to the box shadow so I can merge the changes for og tags!

shubhampandeyhld:
okay..

assets/css/index.css
```diff
@@ -121,6 +121,10 @@ section {
   margin-bottom: 1em;
   font-size: 1.5rem;
 }
+.button-container .sbtn :hover{
```
shahednasser, line 124: Please remove this as well. I just need the changes in `index.html`

shubhampandeyhld:
Hey I have just added the default box-shadow,  so now you can merge the og tags.

shahednasser (changes requested):

assets/css/index.css
shubhampandeyhld: yes , I am removing it

shahednasser:
Great! Thank you!

@all-contributors please add @shubhampandeyhld for code and ideas

PR #67, opened 2020-09-10
Pink button variations

yashgupta18:
Issue:


shahednasser:
Thank you for the great work!

@all-contributors please add @yashgupta18 for code, ideas and example

PR #78, opened 2020-09-11
remove box shadow on click

isonnymichael:
adding box-shadow none on sbutton css selector

Issue:


shahednasser:
Change should only be made to `_click.less`

Issue #42, opened 2020-09-07
Click button animation style


Pratyush-Saxena:
**CHANGE IN CLICK BUTTON STYLE**
I guess Click button that u have made will look much better if it would be like
this one :
![click_button](https://user-images.githubusercontent.com/52444607/92356284-7a767180-f103-11ea-9dae-f0bbfefa3f22.gif)
rather than this one :
![click_2](https://user-images.githubusercontent.com/52444607/92356325-82361600-f103-11ea-85e2-7afdbcc1ceae.gif)

**Additional notes**

shahednasser:
Good idea. Could you work on it?

Pratyush-Saxena:
Yeah sure!

shahednasser:
Resolved by #87

PR #87, opened 2020-09-11
issue #42 resolved

Pratyush-Saxena:
@shahednasser   #42 resolved


shahednasser:
Please add lessc to devDependencies instead of dependencies

Pratyush-Saxena:
> Please add lessc to devDependencies instead of dependencies

 How should i do that?

src/components/_basic.less
```diff
@@ -27,6 +27,7 @@
   &:active,
   &:focus {
     outline: none;
+    box-shadow: 0px 5px 0px @lightGray;
```
shahednasser, line 30: What is this for?

src/components/animated/_click.less
```diff
@@ -2,7 +2,8 @@
   box-shadow: 0px 5px 0px @lightGray;
 
   &:active {
-    box-shadow: 0px 2px 0px @lightGray;
+    transform: translateY(3px);
+    box-shadow: 0px 5px 0px @lightGray;
```
shahednasser, line 6: Add another rule for focus. We don't want the default box-shadow to appear when "Click buttons" are clicked.

shahednasser (changes requested):

shahednasser:
Run `npm remove lessc` then `npm install lessc --save-dev`  
Also, I have tested it there

src/components/animated/_click.less
Pratyush-Saxena: ok then shall i just remove box-Shadow property or anything else is needed to be done?

src/components/_basic.less
Pratyush-Saxena: Actually I thought it would be better..if those buttons get back to their style...as soon as we hover-out the pointer....after clicking....bt its not working in the same way......we have to click some where else in the screen to make those button style back to their own after clicking

src/components/animated/_click.less
shahednasser: change it to the following:
```
  &:active, &:focus {
    box-shadow: 0px 2px 0px @lightGray;
  }

&:active {
    transform: translateY(3px);
}
```

src/components/_basic.less
shahednasser: Please remove any changes to files that are not related to the click button specifically. Changes tp `src/components/_basics.less` can affect all buttons

src/components/animated/_click.less
Pratyush-Saxena: ok

src/components/animated/_click.less
```diff
@@ -1,8 +1,13 @@
 .click-btn {
   box-shadow: 0px 5px 0px @lightGray;
 
+  &:focus {
+    box-shadow: 0px 5px 0px @lightGray;
+  }
+
   &:active {
```
shahednasser, line 8: You still haven't made the change I requested regarding `&:active` and `&:focus` please go check it again and add it.

package.json
```diff
@@ -17,7 +17,9 @@
     "concurrently": "^5.3.0",
     "less": "^3.12.2",
     "less-watch-compiler": "^1.14.6",
+    "lessc": "^1.0.2",
     "onchange": "^7.0.2",
     "prettier": "^2.1.1"
-  }
+  },
+  "dependencies": {}
```
shahednasser, line 24: Please manually remove this line as it is not important anymore.

shahednasser (changes requested):

package.json
Pratyush-Saxena: ok

src/components/animated/_click.less
Pratyush-Saxena: actually this time i have made changes ...to work those buttons in intended manner...without writing any extra code

Pratyush-Saxena (commented):
reviewed

shahednasser:
Great thank you!

@all-contributors please add @Pratyush-Saxena for code and bug

Issue #88, opened 2020-09-11
Replace Download link text with icon


shahednasser:
Replace "Download" in the navbar with an icon. You can use icons from Font awesome as it is already included in the website.

PR #90, opened 2020-09-11
Replaced download text link with icon [Issue #88]

cosay:
Replaced "Download" in the navbar with an font awesome _fa-file-download_ icon.

Issue: #88


shahednasser (changes requested):
Thank you for contributing! Can you change the icon to the [download](https://fontawesome.com/icons/download?style=solid) icon instead? I think it would look better this way

cosay:
Done. It looks better this way 👍

shahednasser:
Awesome, thanks!

@all-contributors please add @cosay for code

PR #93, opened 2020-09-12
Add copied tooltip in How to Use

isonnymichael:
Adding css class script-copy, and embed to code-block-container, then adding function jquery script-copy.

Issue:


assets/js/index.js
```diff
@@ -117,6 +117,25 @@ $(document).ready(function () {
     setTimeout(() => $(this).tooltip("hide"), 1500);
   });
 
+  $(".script-copy").tooltip({
+    title: "Copied",
+    trigger: "click",
+    placement: "top",
```
shahednasser, line 123: Change this to bottom

isonnymichael: ok

isonnymichael:
I add div to wrap the section script and embed class script-copy on it, I recommit because I forgot to delete the last script-copy on last code-block-container

Issue #115, opened 2020-09-13
Change color of yellow-btn for Fill Buttons


shahednasser:
Currently, fill buttons of all types (fill button up, down, etc...) have text color set to `#333`, change it to the same yellow color as the button when it's not hovered, and then on hover use the `#333`. 
Basically, when it's not hovered, it should look like outline button.

cosay:
I take this 😃

shahednasser:
Great!

PR #116, opened 2020-09-13
Do not set dark text on fill color btn [Issue #115]

cosay:
Do not set dark text on fill color btn (added rule in .less file).
Issue #115


shahednasser:
Thank you @cosay !

PR #117, opened 2020-09-14
Caseycling

caseycling:
This pull request closes the sidebar when a link is clicked in small devices.

Issue: Close sidebar on clicking on link in small devices

This is not complete. Lines 144 to 158 add a class 'sm-device-link' to the attributes in screens smaller than 600 pixels and remove it in screens greater than 600 pixels. Lines 161 to 168 add a click handler to the sm-device-link class to close the sidebar. However when the screen is larger, the animation on the sidebar still persists. I checked in the dev tools and the class sm-device-link is being added and removed depending on screen size so I'm not sure why the function is still being called on the click handler.


shahednasser:
Hi @caseycling thank you for taking the time to contribute.

So the reason the click event is not triggered for `.sm-device-link` elements is because the event is added before the element is created. When you add an event listener on click, it only listens to the elements that currently have the class `.sm-device-link` at the time the event listener being added. Later on, if you add the class `.sm-device-link` to an element, the event handler would not be triggered because it didn't have the class before and so the event listener wasn't attached to it. A way to fix this is by doing the following: 
```
$('body').on('click', '.sm-device-link`, function(){...});
```

Here, you're attaching the event handler to the `body` but it will only work to child elements that have the class `.sm-device-link` that way even if you add the class to the element later on, the event listener will be listening to events on that element.

However, I don't think a fix like this is necessary. You can just add a click event to `.sidebar a` and in that event if the screen is less than 768px, you do the work in the event handler you're currently using. I prefer this fix.

So please, make the changes accordingly and let me know if you need any help.

caseycling:
Okay, thank you! I'll make those changes and then send another pr

PR #129, opened 2020-09-16
Removed outline for hamburger menu & close buttons

chelsi-001011:


shahednasser (changes requested):
Please remove any changes in index.html as it is not required. We just need changes in `assets/css/index.css`

chelsi-001011:
> Please remove any changes in index.html as it is not required. We just need changes in `assets/css/index.css`
yes ill do so

chelsi-001011:
I have removed the changes in another commit. @shahednasser

assets/css/index.css
```diff
@@ -162,6 +162,11 @@ hr {
   vertical-align: sub;
 }
 
+.navbar .sidebar-toggler{
```
shahednasser, line 165: This works only for the toggler, please also add a case for `.close-sidebar` with it

chelsi-001011: okay

chelsi-001011: @shahednasser I have fixed the changes

assets/css/index.css
```diff
@@ -162,6 +162,11 @@ hr {
   vertical-align: sub;
 }
 
+.navbar .sidebar-toggler, .close-sidebar{
+    outline:none;
```
shahednasser, line 166: It seems like on `.close-sidebar:active` this is overridden by another rule, so please add `!important` to `outline: none`

shahednasser (changes requested):

assets/css/index.css
chelsi-001011: If by overridden you mean that on clicking the .close-sidebar button, the section slides to the left and instantly gets back in the original position. Is it supposed to happen? @shahednasser

shahednasser: No, I mean that on click the `.close-sidebar` button becomes focused and due to CSS rules made by bootstrap, an outline is added. To make sure our CSS rules are applied instead of bootstrap's, add `!important` to the rule

chelsi-001011: Ill add the changes

shahednasser:
Awesome! Thank you for your work.

@all-contributors please add @chelsi-001011 for code

chelsi-001011:
> Awesome! Thank you for your work.
> 
> @all-contributors please add @chelsi-001011 for code

Thank you!

Issue #134, opened 2020-09-16
Make "display: inline-block;" for a.sbtn


shahednasser:
Currently add sbuttons designs are aimed towards `<button>` elements. To add support to links as well, we need to add `display: inline-block;` for `<a>` elements that have the class `.sbtn`. Changes should be made in `src/components/_basic.less` because it should be applied to all button types

adityasinghbaghel:
Hey!  I want  to  contribute and Fix the issue . . .

PR #136, opened 2020-09-16
Added "display: inline-block;" for a.sbtn

adityasinghbaghel:
Issue:
#134


src/components/_basic.less
```diff
@@ -35,6 +35,7 @@ a.sbtn {
   &:hover,
```
shahednasser, line 35: remove `&:hover, &:active` it should be on `a.sbtn` and it'll be applied for all states as well.

shahednasser (changes requested):
You need to compile your code by running `npm run compile` in order for the changes to take effect.

src/components/_basic.less
adityasinghbaghel: ok

shahednasser (changes requested):
Are you sure you compiled the less files? If you ran the compile command which is `npm run compile`, `dist/sbuttons.min.css` would've changed too. Since it hasn't changed, no changes can be seen on the website. If you're having trouble with this, please check the [contribution guide](https://github.com/shahednasser/sbuttons/blob/master/CONTRIBUTING.md) for more details

adityasinghbaghel:
Sorry , it was my mistake I forgot to run ( npm run watch)

shahednasser:
thank you!

@all-contributors please add @adityasinghbaghel for code and bug

Issue #132, opened 2020-09-16
Add "How to Use" section to sidebar menu


shahednasser:
Add a link to how to use section in the sidebar menu

lee00286:
Can I work on this issue?

shahednasser:
Yes!

PR #137, opened 2020-09-17
Added How to Use section to sidebar

lee00286:
Edited `buttons-examples.js` to add How to Use section to sidebar

Issue: #132


shahednasser (changes requested):
Instead of adding it to `buttons-examples.js` please add it directly to `index.html` below `.close-sidebar`. Let's keep this file for buttons examples only

shahednasser:
Great thank you!

Issue #114, opened 2020-09-13
Change color of .button-caption-sub in dark mode


shahednasser:
Change color of `.button-caption-sub` to `#afafaf` in dark mode. You can refer to `assets/css/index.html` to figure out how to style elements in dark mode.

nileshanand16:
can I take this issue?

shahednasser:
Yes of course!

nileshanand16:
}

.button-caption-sub {
  color: #afafaf;
  font-family: monospace;
  cursor: pointer;
}

shahednasser:
You need to fork the repository, make changes to the code then create a pull request. Read more about it in the [contributing guidelines](https://github.com/shahednasser/sbuttons/blob/master/CONTRIBUTING.md)

nileshanand16:
DONE

shahednasser:
Hi @nileshanand16 you have pushed a commit to your repository that you forked from this. You need to create a pull request to merge your code. Please check the [contribution guidelines](https://github.com/shahednasser/sbuttons/blob/master/CONTRIBUTING.md) for more details

HADES-01:
Can I work this issue ????

shahednasser:
Yes @HADES-01  go ahead!

PR #147, opened 2020-09-19
.button-caption-sub in dark mode #114 updated

nileshanand16:
Issue:


shahednasser (changes requested):
Great work but we need it applied only in dark mode. Don't change the previously assigned color, add a new rule that's only applied in dark mode. If you need, you can check other rules in this file to see how it's done.

Issue #85, opened 2020-09-11
Close sidebar on clicking outside of it


shahednasser:
Currently, only clicking the close button closes the sidebar. Add a click event that closes the sidebar by clicking outside of it.

Pratyush-Saxena:
I would like to take this up @shahednasser

shahednasser:
Sure!

Pratyush-Saxena:
@shahednasser ...can you pls tell the condition that checks ..whether the sidebar is open or not

shahednasser:
the sidebar should have class `.animate__slideInLeft` and not have the class `.hide-sidebar` when it is opened

Pratyush-Saxena:
> .animate__slideInLeft

it isn't working...i am trying something like this ->
```
$(window).on("click",function (event) {
    event.stopPropagation();
    let sidebar = $(".sidebar");
    if(event.target!=sidebar && sidebar[0].classList.contains(".animate__slideInLeft") && (!sidebar[0].classList.contains(".hide-sidebar"))){
      sidebar.addClass("animate__slideOutLeft");
    window.setTimeout(function () {
      sidebar.toggleClass("hide-sidebar");
      sidebar.removeClass("animate__slideOutLeft");
    }, 1000);
    }
  }); 
```

shahednasser:
You can't assign `$(".sidebar")` to `sidebar` and then use it as a JS DOM element in the following code. Please use jquery functions (for example, instead of `sidebar[0].classList.contains(".animate__slideInLeft")` use `sidebar.hasClass("animate__slideInLeft")`

Pratyush-Saxena:
> You can't assign `$(".sidebar")` to `sidebar` and then use it as a JS DOM element in the following code. Please use jquery functions (for example, instead of `sidebar[0].classList.contains(".animate__slideInLeft")` use `sidebar.hasClass("animate__slideInLeft")`

that too isn't  working

shahednasser:
In your code, it also says `$(window).on('click')...` try changing `window` to `"body"`

PR #153, opened 2020-09-20
issue #85 resolved

Pratyush-Saxena:
@shahednasser i have resolved issue #85 ...pls review this pr

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -156,8 +156,9 @@ $(document).ready(function () {
   });
 
   $("body").click(function(event) {
-    let sidebar = $(".sidebar") ;
-    if( $(window).width() <= 640 && event.target.className.split(" ",1)!="sidebar" && event.target.className.split(" ",1)!="fas" && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
+    var sidebar = $(".sidebar") ;
+    var targetElement = $(event.target);
+    if( $(window).width() <= 767 && !targetElement.hasClass("sidebar") && !targetElement.hasClass("fas") && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
       sidebar.addClass("animate__slideOutLeft");
       window.setTimeout(function () {
         sidebar.toggleClass("hide-sidebar");
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -158,7 +158,7 @@ $(document).ready(function () {
   $("body").click(function(event) {
     var sidebar = $(".sidebar") ;
     var targetElement = $(event.target);
-    if( $(window).width() <= 767 && !targetElement.hasClass("sidebar") && !targetElement.hasClass("fas") && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
+    if( $(window).width() <= 767 && !targetElement.hasClass("sidebar") && !targetElement.parents('.sidebar').length && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
       sidebar.addClass("animate__slideOutLeft");
       window.setTimeout(function () {
         sidebar.toggleClass("hide-sidebar");
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -136,8 +136,9 @@ $(document).ready(function () {
     setTimeout(() => $(this).tooltip("hide"), 1500);
   });
 
-  $(".sidebar-toggler").on("click", function () {
+  $(".sidebar-toggler").on("click", function (e) {
     let sidebar = $(".sidebar");
+    e.stopPropagation()
     sidebar.toggleClass("hide-sidebar");
   });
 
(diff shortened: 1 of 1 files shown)


assets/js/index.js
```diff
@@ -154,6 +154,16 @@ $(document).ready(function () {
     }
   });
 
+  $("body").click(function(event) {
+    let sidebar = $(".sidebar");
```
shahednasser, line 158: add an if condition that checks:
1. window width is less than 768px
2. sidebar has class `animate__slideInLeft`

then put the rest of the code of the function inside it.

assets/js/index.js
```diff
@@ -154,6 +154,16 @@ $(document).ready(function () {
...
+    if(event.target.className.split(" ",1)!="sidebar" && event.target.className.split(" ",1)!="fas" && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
```
shahednasser, line 159: before this if condition, add a new variable `eventTarget = $(event.target)` and then in the if condition use functions like `hasClass` instead of split

shahednasser (changes requested):
Also, please add the following:

1. On line 139 `$(".sidebar-toggler").on("click", function () {` add e to  the function like so: `$(".sidebar-toggler").on("click", function (e) {`
2. On line 141, after `sidebar.toggleClass("hide-sidebar");` add `e.stopPropagation()`

This will stop the new event from executing when clicking the hamburger menu to open the sidebar. Sometimes clicking it is opening and closing fast.

assets/js/index.js
Pratyush-Saxena: i found ```hasclass``` not working properly so i used ```split```

Pratyush-Saxena: > before this if condition, add a new variable `eventTarget = $(event.target)` and then in the if condition use functions like `hasClass` instead of split

we can't do ```eventTarget = $(event.target)  ``` because on doing so it is throwing an error ``` split() not define```

Pratyush-Saxena (commented):
@shahednasser  i have reviewed and commiting with all the possible changes you requested

assets/js/index.js
```diff
@@ -154,6 +155,16 @@ $(document).ready(function () {
     }
   });
 
+  $("body").click(function(event) {
+    let sidebar = $(".sidebar") ;
```
shahednasser, line 159: Change let to var for browser compatibility issues.

assets/js/index.js
```diff
@@ -154,6 +155,16 @@ $(document).ready(function () {
...
+    if( $(window).width() <= 640 && event.target.className.split(" ",1)!="sidebar" && event.target.className.split(" ",1)!="fas" && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
```
shahednasser, line 160: There's still this change you haven't made regarding `event.target`. Here's how it should look like:
```
var targetElement = $(event.target);
```
This should be before the if.  

Then use it in the if like this: `!targetElement.hasClass("sidebar")`

shahednasser, line 160: Why is it 640 instead of 767?

shahednasser (changes requested):

assets/js/index.js
Pratyush-Saxena: because you specified here
``` $(".scroll-top").click(function () 
    let scrollSpeed = 100; /* Default */
    if ($(window).width() <= 640)
      scrollSpeed = 210; /* For small (mobile) Screens */
```

assets/js/index.js
Pratyush-Saxena: > > before this if condition, add a new variable `eventTarget = $(event.target)` and then in the if condition use functions like `hasClass` instead of split
> 
> we can't do `eventTarget = $(event.target) ` because on doing so it is throwing an error ` split() not define`

i tried this before but it is throwing an error..!!

assets/js/index.js
Pratyush-Saxena: i tried this before  before ....but it is throwing an error ```split() is not defined```

shahednasser: We don't need split anymore in this case. Instead of  `event.target.className.split(" ",1)!="sidebar"` it will just be `!targetElement.hasClass("sidebar")`

shahednasser: No this is for a different issue. Please change it to 767, as it is the size of the screen where sidebar appears by clicking on the hamburger menu.

Pratyush-Saxena: > We don't need split anymore in this case. Instead of `event.target.className.split(" ",1)!="sidebar"` it will just be `!targetElement.hasClass("sidebar")`

I'll surely do that but i doubt it may cause error anytime...

assets/js/index.js
```diff
@@ -154,6 +155,17 @@ $(document).ready(function () {
     }
   });
 
+  $("body").click(function(event) {
+    var sidebar = $(".sidebar") ;
+    var targetElement = $(event.target);
+    if( $(window).width() <= 767 && !targetElement.hasClass("sidebar") && !targetElement.hasClass("fas") && sidebar.hasClass("animate__slideInLeft") && (!sidebar.hasClass("hide-sidebar"))){
```
shahednasser, line 161: Replace `!targetElement.hasClass("fas")` with `!targetElement.hasClass("sidebar") && !targetElement.parents('.sidebar').length`. This ensures that when clicking on the sidebar or any of its child elements, this event listener will not continue.

shahednasser (changes requested):
This works fine but it's a little buggy. After debugging it a lot, I have requested a few changes to make sure it works perfectly.

Pratyush-Saxena (commented):
@shahednasser  reviewed the changes you want please check out the last commit

shahednasser:
Thank you @Pratyush-Saxena for the great work!

@all-contributors please add @Pratyush-Saxena for code

Issue #158, opened 2020-09-21
Download Button on website


reveurguy:
**Describe the bug**
The download button on website opens a page with all the code. Instead it should download a file with all the code.
**Expected behavior of what should have happened (if applicable)**

**Screenshots (if applicable)**

**Additional notes**

shahednasser:
Hi @reveurguy thank you for submitting this. I'm actually aware of that but as of now it's fine. I'll add it to enhancements though as it is something that's worth working on.

reveurguy:
also i want to suggest to use a lighter color for dark mode, like grey or black with some opacity or a lighter black color, so that the black buttons are clearly visible.

shahednasser:
You can submit a PR with what you mean and we'll see if it's better

PR #166, opened 2020-09-23
Fixed Download Button on website

aymenhamada:
The download button now download a file with all the code instead of opening the raw github page.

<a download / wasn't working as raw github page isn't considered like a file , so i came up with using JS & fetch API.

Issue: #158

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -98,6 +98,24 @@ $(document).ready(function () {
     setTimeout(() => $element.tooltip("hide"), 1500);
   }
 
+  function downloadGithubCssRaw(){
+    const link = "https://raw.githubusercontent.com/shahednasser/sbuttons/master/dist/sbuttons.min.css";
+    fetch(link)
+      .then(resp => resp.blob())
+      .then(blob => {
+        const url = window.URL.createObjectURL(blob);
+        const a = document.createElement('a');
+        a.style.display = 'none';
+        a.href = url;
+        a.download = "sbuttons.min.css";
+        document.body.appendChild(a);
+        a.click();
+        window.URL.revokeObjectURL(url);
+        document.body.removeChild(a);
+  })
+  .catch((err) => console.log(err));
+  }
+
   var copiedTooltipOptions = {
     title: "Copied",
     trigger: "click",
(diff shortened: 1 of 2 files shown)


assets/js/index.js
```diff
@@ -98,6 +98,24 @@ $(document).ready(function () {
     setTimeout(() => $element.tooltip("hide"), 1500);
   }
 
+  function downloadGithubCssRaw(){
+    const link = "https://raw.githubusercontent.com/shahednasser/sbuttons/master/dist/sbuttons.min.css";
```
shahednasser, line 102: This is a good approach, however, fetch isn't compatible with all browsers. Please surround `fetch` code with `if (window.fetch)` to check it's supported in the browser and in the else do something like `window.location.href = "https://raw.githubusercontent.com/shahednasser/sbuttons/master/dist/sbuttons.min.css"`  

This way we can support all devices. Also, please replace `const link` with `var link` for browser compatibility as well. Inside the fetch is fine because we already know it will work because of the if condition.  

If there's anything not clear, please let me know.

shahednasser (changes requested):

assets/js/index.js
shahednasser: There's also a link that leads to the CSS file in "How to Use" section. Can you also attach the event listener to it as well?

aymenhamada: Sure, i'll work on that

index.html
```diff
@@ -86,7 +86,7 @@ <h1 class="section-header">How to Use</h1>
             <div class="instructions">
               <div class="text">
                 You can download the CSS file
-                <a href="https://github.com/shahednasser/sbuttons/blob/master/dist/sbuttons.min.css" class="cssLink"
+                <a href="#" id="downloadGithubRawHowTo" class="download-link" class="cssLink"
```
shahednasser, line 89: can you remove `class="download-link"` as it already has class `cssLink`

shahednasser (changes requested):

index.html
aymenhamada: Oh yeah sorry for the mistake. i'll fix that

shahednasser:
Awesome work! Thank you.

@all-contributors please add @aymenhamada for code and bug

Issue #83, opened 2020-09-11
Add overlay when sidebar opens on small devices


shahednasser:
When the sidebar opens on small devices, add an overlay that darkens a little the rest of the page.

shubhampandeyhld:
Please assign me this issue also.

shahednasser:
ok I assigned you

shahednasser:
Hello @shubhampandeyhld is there any update on this?

shubhampandeyhld:
Sorry I am bit busy these days.
No I can't update .

shahednasser:
I understand it's not a problem. If by the end of this weekend you still find it hard to work on these issues let me know so maybe we can give them to someone else.

shahednasser:
Unassigning to give others a chance

JLambertazzo:
Hey, I'd like to work on this! @shahednasser

shahednasser:
Hi @JLambertazzo go ahead!

PR #180, opened 2020-09-24
Added overlay

JLambertazzo:
Issue: #83 Add overlay when sidebar opens on small devices
I added a div as an overlay to darken everything but the sidebar on smaller devices.


assets/css/sidebar.css
```diff
@@ -97,4 +102,13 @@
...
     z-index: 100;
   }
+  
+  #shade.shown {
+    width: 100vw;
+    height: 100vh;
+    z-index: 9;
+    position: fixed;
+    background-color: black;
+    opacity: 30%;
```
shahednasser, line 112: Change opacity to 50%

assets/css/sidebar.css
```diff
@@ -57,6 +57,11 @@
   color: #fff;
 }
 
+#shade {
+  opacity: 0%;
+  transition: opacity 2s;
```
shahednasser, line 62: change transition duration to 0.2s instead of 2

shahednasser (changes requested):
It all works perfectly! However, we just added a new feature where clicking outside the sidebar can closes it. Can you update the code, and toggle the class "shown" in that event as well? If you need any assistance in doing this please let me know.

JLambertazzo:
Done!

shahednasser (changes requested):
It all works perfectly! But I just noticed that when clicking on one of the links in the sidebar, the overlay is still there so it needs to be added to that event. I think we need to make closing the sidebar a function itself so that we don't have so much repeated code. For now, please add the toggle to that event as well. If you find other events in the file that closes the sidebar as well, please add it there.

JLambertazzo:
Oops, fixed that!

shahednasser:
Amazing work! Thank you.

@all-contributors please add @JLambertazzo for code

PR #188, opened 2020-09-25
update BUTTON_IDEAS.md

devSouvik:
new idea added for icon buttons icons will shirk and enlarge upon hover


shahednasser:
Can you please move the idea to the "icons" category, as it will use an icon in it (even if it's animated)? The reason behind this is that icon buttons will use font awesome and we want them grouped together basically.

devSouvik:
ok..done

shahednasser:
Great! Thank you so much.

@all-contributors please add @devSouvik for idea

PR #190, opened 2020-09-25
Plus button

zilyukina:
Button with plus icon added


shahednasser:
Thank you @zilyukina for the great work!

@all-contributors please add @zilyukina for code

PR #197, opened 2020-09-25
watch:all & lite-server added

chaitanya4vedi:
watch:all for less and sass and lite-server added

1. First Step
To run or work on the the project locally we need dependencies mentioned on package.json as it is required by the project.
`npm install` in the terminal will install all the required dependencies.
you might be prompted to run `npm audit fix` at the end. This will automatically fix the vulnerabilities if any.

2. Second Step
To run the project locally we do
`npm start`
this will trigger all the commands like `npm run build`, `npm run dev`, `lite-server`,etc. mentioned in the `start` script and the project will eventually run in your default browser.


.gitignore
```diff
@@ -1 +1,5 @@
 node_modules/
+.cache
+.idea
+.vs
+dist
```
shahednasser, line 5: dist shouldn't be in gitignore please remove that

package.json
```diff
@@ -20,5 +26,12 @@
     "lessc": "^1.0.2",
     "onchange": "^7.0.2",
     "prettier": "^2.1.1"
+  },
+  "dependencies": {
```
shahednasser, line 30: Can you change these to "devdependencies" as they are required for development purposes and not for sButtons to work on its own?

shahednasser (changes requested):
Can you explain what each of the new commands does? We need to add the explanation to the contribution guide too to help contributors understand what command they need.

.gitignore
chaitanya4vedi: But dist is created every time we run the project. It is not recommended to exclude it from git-ignore. Also there will be be a long dist files to be committed together with every new commit if it is not ignored. What we can do is to add `/dist` to make sure only top level dist is ignored.

package.json
chaitanya4vedi: Ok

.gitignore
shahednasser: Dist is where the code for `sbuttons.min.css` resides in, which we need. Do you have any other idea of where it should be?

chaitanya4vedi: I see. No i just knew that it should be in .git-ignore. But in our case i see its not recommended.

package.json
```diff
@@ -6,11 +6,18 @@
...
     "compile": "lessc src/sbuttons.less dist/sbuttons.css && npm run prettier && npm run clean-css",
     "watch": "concurrently \"less-watch-compiler src dist\" \"onchange -v dist/sbuttons.css -- npm run clean-css \" ",
     "prettier": "prettier --write \"src/**/*.less\" \"dist/sbuttons.css\" \"assets/**\" ",
+    "lite": "lite-server",
     "check-prettier": "prettier --check \"src/**/*.less\" \"dist/sbuttons.css\" ",
-    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css"
+    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css",
+    "watch:all": "parallelshell 'npm run watch:scss' 'npm run watch:less' 'npm run lite'",
+    "watch:scss": "onchange 'css/*.scss' --npm run scss",
+    "watch:less": "onchange 'css/*.less' --npm run less"
```
shahednasser, line 20: The command for "watch:less" should be `onchange -v src/*.less -- npm run compile`

package.json
```diff
@@ -19,6 +26,12 @@
...
-    "prettier": "^2.1.1"
-  }
+    "prettier": "^2.1.1",
+    "lite-server": "^2.5.4",
+    "node-less": "^1.0.0",
+    "node-sass": "^4.14.1",
+    "parallelshell": "^3.0.2",
+    "parcel-bundler": "^1.12.4"
+  },
+  "dependencies": {}
```
shahednasser, line 36: remove this

package.json
```diff
@@ -6,11 +6,18 @@
   "author": "Shahed Nasser",
   "license": "ISC",
   "scripts": {
+    "build": "parcel build index.html",
+    "dev": "parcel index.html --open && lite-server",
+    "start": "npm run build && npm run dev && npm run watch:all",
```
shahednasser, line 11: the "npm run watch:all" will never work, as the execution is stuck in "npm run dev" so maybe you can use "parallelshell" here for "npm run dev" and "npm run watch"

shahednasser (changes requested):
The lite-server works great. However there are some issues with the watch command, you can see the fix for it in the review below.  

Also, I don't think it's necessary to add a watch for scss, as we don't use it in this project. Can you remove "watch:scss" and its dependencies? As we only will need watch for less, remove all watch commands (including the one that was already there before) and keep "watch:less" but rename it to "watch". Again, make sure to remove all dependencies that we will not need after doing this. We don't want to add any unnecessary dependencies.

shahednasser:
I also forgot to add that we need to add details about what to run for development in the [contributing guidelines](https://github.com/shahednasser/sbuttons/blob/master/CONTRIBUTING.md). Let me know if you can do it or not.

shahednasser:
Hello @chaitanya4vedi we now have switched hosting sbuttons.min.css using [Statically](https://statically.io/), so the compilation now will only generate "dist/sbuttons.css", while the minification is done by Statically.

chaitanya4vedi:
> Hello @chaitanya4vedi we now have switched hosting sbuttons.min.css using [Statically](https://statically.io/), so the compilation now will only generate "dist/sbuttons.css", while the minification is done by Statically.

Great!

package.json
```diff
@@ -6,11 +6,15 @@
...
   "scripts": {
+    "start": "npm run watch:all",
     "compile": "lessc src/sbuttons.less dist/sbuttons.css && npm run prettier && npm run clean-css",
     "watch": "concurrently \"less-watch-compiler src dist\" \"onchange -v dist/sbuttons.css -- npm run clean-css \" ",
     "prettier": "prettier --write \"src/**/*.less\" \"dist/sbuttons.css\" \"assets/**\" ",
+    "lite": "lite-server",
     "check-prettier": "prettier --check \"src/**/*.less\" \"dist/sbuttons.css\" ",
-    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css"
+    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css",
+    "watch:less": "onchange -v src/*.less -- npm run compile",
```
shahednasser, line 16: Please remove the "watch" command and rename "watch:less" to just watch. Or rename "watch:all" to "watch". We don't need the previous watch command anymore.

package.json
```diff
@@ -6,11 +6,15 @@
...
+    "watch:all": "parallelshell 'npm run watch:less' 'npm run lite'"
```
shahednasser, line 17: parallelshell is not throwing an error for some reason. Maybe use concurrently?

shahednasser (changes requested):

shahednasser (changes requested):
Great. One final request is to add concurrently to the dev dependencies, as it has been removed previously.

Also, after the latest change regarding sbuttons.min.css, now we have a problem which is that during development, the contributor can't see the changes they're making, since the script in head refers to the sbuttons.css in this repository. Do you think we should add in `<head>` of index a link to "dist/sbuttons.css" or is there a better idea?

shahednasser:
I have currently added "dist/sbuttons.css" instead of min for development purposes. We don't want to cause any confusion to our contributors. If you have any better idea though please let me know.

chaitanya4vedi:
concurrently was there but it is not showing up. I reinstalled the latest version but it is still not showing up

package.json
```diff
@@ -6,18 +6,23 @@
...
+    "start": "npm run watch",
     "compile": "lessc src/sbuttons.less dist/sbuttons.css && npm run prettier && npm run clean-css",
-    "watch": "concurrently \"less-watch-compiler src dist\" \"onchange -v dist/sbuttons.css -- npm run clean-css \" ",
     "prettier": "prettier --write \"src/**/*.less\" \"dist/sbuttons.css\" \"assets/**\" ",
+    "lite": "lite-server",
     "check-prettier": "prettier --check \"src/**/*.less\" \"dist/sbuttons.css\" ",
-    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css"
+    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css",
+    "watch:less": "onchange -v src/*.less -- npm run compile",
+    "watch": "concurrently 'npm run watch:less' 'npm run lite'"
```
shahednasser, line 16: It seems concurrently required double quotations instead of single quotations so please replace the command with `concurrently \"npm run watch:less\" \"npm run lite\"`

package.json
```diff
@@ -6,18 +6,23 @@
...
   "scripts": {
+    "start": "npm run watch",
     "compile": "lessc src/sbuttons.less dist/sbuttons.css && npm run prettier && npm run clean-css",
-    "watch": "concurrently \"less-watch-compiler src dist\" \"onchange -v dist/sbuttons.css -- npm run clean-css \" ",
     "prettier": "prettier --write \"src/**/*.less\" \"dist/sbuttons.css\" \"assets/**\" ",
+    "lite": "lite-server",
     "check-prettier": "prettier --check \"src/**/*.less\" \"dist/sbuttons.css\" ",
-    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css"
+    "clean-css": "cleancss -o dist/sbuttons.min.css dist/sbuttons.css",
+    "watch:less": "onchange -v src/*.less -- npm run compile",
```
shahednasser, line 15: Change watch:less to `onchange -v src/**/*.less -- npm run compile` to detect changes inside nested files in src

shahednasser (changes requested):
With these two changes, I tested it and it works great! You've worked hard thank you

package.json
chaitanya4vedi: I see. Ok

package.json
chaitanya4vedi: Done

package.json
chaitanya4vedi: Done

package.json
chaitanya4vedi: Done

chaitanya4vedi:
> With these two changes, I tested it and it works great! You've worked hard thank you

Thanks! Indeed you did a harder job . 
Looking forward to contribute more

shahednasser:
Great work thank you so much!

@all-contributors please add @chaitanya4vedi for code, ideas, and tools

Issue #96, opened 2020-09-12
Add button ideas


shahednasser:
There's now a new page for [button ideas](https://github.com/shahednasser/sbuttons/blob/master/BUTTON_IDEAS.md) where you can add ideas that you would like to see implemented. If you have any ideas that you are not able to implement for any reason, please add them there.

If you would like to check button ideas to implement, make sure to look there, and follow the [contibuting guidelines](https://github.com/shahednasser/sbuttons/blob/master/CONTRIBUTING.md) to implement the idea and check it off.

esakki-selvaraj:
We can add neon light buttons using the shadow attribute

shahednasser:
Great! Can you please edit the [button ideas](https://github.com/shahednasser/sbuttons/blob/master/BUTTON_IDEAS.md) page to include your idea so others can see it! You will also be added as a contributor to this project.

esakki-selvaraj:
sure will do it

Maitryee-Choudhary:
Hey,can I work on this project?

shahednasser:
Sure @Maitryee-Choudhary ! This issue has the label "Always Open" and is open for everyone

alinasahoo:
Hey, we can have a button as of like when we click on it, the color changes into another color and says success if passed all the test cases. 

I have recently joined GitHub and am unaware of where to start contributing. I am a bit inclined towards front-end development so it would be helpful if you would guide me a bit. Also, are there any more issues where I can work as of now?

shahednasser:
Hi @alinasahoo looks like it's a good idea. There are two ways you can contribute with this idea:

1. Submit the idea by editing the [button ideas](https://github.com/shahednasser/sbuttons/blob/master/BUTTON_IDEAS.md) file
2. Implement the button if you think you can do it.

If you need any assistance you can always ask! But before contributing, please check out the [contributing guidelines](https://github.com/shahednasser/sbuttons/blob/master/CONTRIBUTING.md) to understand more!

alinasahoo:
Okay, as of now. I only had this idea. Will look into the implementation in a while if I can do it. Meanwhile, I will add this idea in the button ideas file so that others can look into it and start contributing. 

Thank you @shahednasser .

abdullah2205:
hi, i have an idea for a button, can it help this project

shahednasser:
@abdullah2205 yes of course! You add add your idea in the button ideas file or implement it right away!

[...]

PR #199, opened 2020-09-25
Neumorphic button idea

vitoriapena:
Adding this button idea that I'd like to implement

Issue: #96


shahednasser:
Great thanks for the help!

@all-contributors please add @vitoriapena for idea

PR #200, opened 2020-09-25
Sbuttons logo

chaitanya4vedi:
A primary logo with five alternatives added. Most of them are transparent and one or two is with a solid background.


shahednasser:
I like the logo. Can we get a favicon of maybe the X in the logo? Not with transparent background though, similar to the star we have now.

shahednasser:
It looks good, but I can't merge it for now as it has changes from the lite server pull request, so we need to resolve that one first. Once we do, I'll merge both. 

In the meantime, can you remove the alternative logo options as we don't need them anymore?

chaitanya4vedi:
As asked I added favicon and logo, removing changes from lite-server pull request. I committed a lot of changes before so it was difficult to revert back to the initial commit so I pulled changes from sbuttons. And then added favicon and logo. All other are latest changes of sbuttons repo.

PR #202, opened 2020-09-25
Add Play Icon Button

aparna202:
added a play button icon.


assets/js/buttons-examples.js
```diff
@@ -131,6 +131,12 @@ $(document).ready(function () {
           classes: "like-btn",
           text: "",
         },
+        {
+          type: "play",
+          label: "play button",
```
shahednasser, line 136: Capitalize the label

src/components/icon/_play.less
```diff
@@ -0,0 +1,34 @@
+.play-btn {
+  &:before {
+    content: "\f152";
```
shahednasser, line 3: Can you replace it with a borderless icon? Its code is f04b

shahednasser (changes requested):
Looks good! I just have these two comments

PR #204, opened 2020-09-25
Header download & documentation buttons (animated click & outline)

nicholedwight:
Fixing this issue: https://github.com/shahednasser/sbuttons/issues/133

I added two buttons to the header under the subtitle text, the download button is an animated click with a white background and black text, the documentation link button has a white background with black text and border.

I also added css to index.css for overriding the button colors to be white (along with comments for other contributors), I'm more than happy to make a permanent white class for all the buttons though!

✨

### DIFF
--- a/index.html
+++ b/index.html
@@ -68,8 +68,8 @@
         <div class="subtitle">
           Simple buttons you can use easily for your next project.
         </div>
-        <a href="#" id="downloadGithubRawButton" class="sbtn click-btn btn-rounded">Download</a>
-        <a href="https://github.com/shahednasser/sbuttons" target="_blank" class="sbtn outline-btn btn-rounded black-btn">Documentation</a>
+        <a href="#" id="downloadGithubRawButton" class="sbtn click-btn btn-rounded yellow-btn">Download</a>
+        <a href="https://github.com/shahednasser/sbuttons" target="_blank" class="sbtn click-btn btn-rounded">Documentation</a>
       </div>
     </header>
     <div class="container mt-5">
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -68,6 +68,8 @@ <h1 class="noselect">sButtons</h1>
         <div class="subtitle">
           Simple buttons you can use easily for your next project.
         </div>
+        <a href="#" id="downloadGithubRawButton" class="sbtn click-btn btn-rounded">Download</a>
+        <a href="https://github.com/shahednasser/sbuttons" target="_blank" class="sbtn outline-btn btn-rounded black-btn">Documentation</a>
```
shahednasser, line 72: I noticed you added a `black-btn` class here, but I don't think it's necessary.

shahednasser (changes requested):
I like your work! I think it would be best if both buttons were click buttons because the first button looks much better. But we don't want them to be the same color, so maybe use `yellow-btn` for one of them but make the other white? Or if you have another color in mind (other than the ones we are already using) you can use it too! Make sure that the color that stands out is the color for the download button, as we need the focus to be on it.

Someone is already working on adding a white button, so hopefully when they're done we'll be able to replace the override with that class!

index.html
nicholedwight: You got it! I'll make these changes :)

shahednasser:
Looks great! Just one final change: I feel like it'll look better if the button's font was larger. Can you change the font-size to 1.3rem?

nicholedwight:
Coming right up!

shahednasser:
Awesome! Thank you so much.

@all-contributors please add @nicholedwight for code

PR #224, opened 2020-09-26
Hide Download button in the navbar when at the top of the page

### LINKED ISSUE #216: Hide "Download" button in the navbar when at the top of the page
When at the top of the page and the header is visible, hide the download link in the navbar. A tip: When the navbar doesn't have the class "scrolling" the download link should be hidden. When it has the class "scrolling", the download link should be visible.

HADES-01:
toggle hide class on download-link in the checkScrollTop function.

Resolves #216

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -384,3 +384,7 @@ button.clipboard:focus {
 .script-copy {
   cursor: pointer;
 }
+
+.hide {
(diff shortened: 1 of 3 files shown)


shahednasser:
Hi @HADES-01 sorry someone is already working on this! Please look at other issues you can contribute to!

Issue #226, opened 2020-09-26
The neon buttons should have more color options rather than just green


HADES-01:
**Describe the solution you'd like**
Add more classes to [_neon.less](https://github.com/shahednasser/sbuttons/blob/master/src/components/animated/_neon.less).
And required examples should be added to the [index.html](https://github.com/shahednasser/sbuttons/blob/master/index.html) as well.

**Additional notes**
Should at least have support for all the current color combinations.

**Currently Working On It**
Yes

shahednasser:
I like the idea. Please work on it!

PR #246, opened 2020-09-26
Added different neon colours effect

HADES-01:
@shahednasser I have added the support for different neon colour combinations.

Resolves: #226


shahednasser:
Love it! Thank you.

@all-contributors please add @HADES-01 for code

Issue #252, opened 2020-09-26
Banner links don't go with the overall aesthetic of the website


HADES-01:
**Describe the bug**
The "Download" and "Documentation" links in the banner of the website don't go with the overall aesthetic of website.
**Screenshots (if applicable)**
Should look like this
![Should look like](https://user-images.githubusercontent.com/52060608/94347255-2f8bb200-0050-11eb-86f6-3c26b1bbb64d.png)

shahednasser:
If you can make a pull request with the change I might look into it. But can you at least make the download button a white color? We need to put focus on the download button.

PR #256, opened 2020-09-26
Changed UI of the Banner

HADES-01:
It looks like this...
Changed

@shahednasser Can you check if its all right or not now?

Resolves: #252

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -388,3 +388,31 @@ button.clipboard:focus {
 .script-copy {
   cursor: pointer;
 }
+
+.banner-link:hover,
+.banner-link:active {
+  color: white;
+  text-decoration: none;
+  background-color: white;
+  color: #333;
+  transition: all 0.3s;
+}
+
+.banner-link {
+  border: 1px solid white;
+  padding: 15px;
+  font-size: 1.3rem;
+  margin-right: 20px;
+  color: white;
+}
+
+.down-link {
+  background-color: white;
+  color: #333;
+}
+
+.down-link:hover,
+.down-link:active {
+  background-color: transparent;
+  color: white;
+}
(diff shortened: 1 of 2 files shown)


shahednasser:
There seems to be a lot of issues due to merging branches. Can you please work it out?

HADES-01:
@shahednasser But it shows me that there are no merge conflicts. Or do you want me to do something about the commits?

shahednasser:
I'm not sure if you've noticed but some files include something like "<<<<<master" or "=====" this is due to resolving merge conflicts in a wrong way.

HADES-01:
@shahednasser Ok, It must have happened because of the git extension on VS code. I will definitely look into it tomorrow.

HADES-01:
@shahednasser I think its fine now.

assets/css/index.css
```diff
@@ -388,3 +388,30 @@ button.clipboard:focus {
...
+  text-decoration: none;
+  background-color: white;
+  color: #333;
+  transition: all 0.3s;
+}
+
+.banner-link {
+  border: 1px solid white;
+  padding: 15px;
+  font-size: 24px;
```
shahednasser, line 404: Change font size to 1.3rem

assets/css/index.css
```diff
@@ -388,3 +388,30 @@ button.clipboard:focus {
...
+}
+
+.down-link {
+  background-color: white;
+  color: #333;
+}
+
+.down-link:hover,
+.down-link:active {
+  filter: brightness(90%);
```
shahednasser, line 416: Instead of changing the brightness, I think it's better if on hover the button becomes transparent and the color changes to white.

HADES-01: @shahednasser I did the required changes.

shahednasser:
Great thank you @HADES-01

PR #259, opened 2020-09-26
Changed headers to Syne Font

hbach24:
Issue: Changed the fonts for the "Download" and "Documentation" links in the header to "Syne"


shahednasser (changes requested):
Instead of inline CSS, can you please put it in `assets/css/index.html` something like 

```
header .sbtn {
font-family: Syne, sans-serif;
}
```

If it requires `!important` then add it

PR #260, opened 2020-09-27
fixing like button animation

isonnymichael:
I decreasse keyframe, and add more duration for animating like button, I try to make it simple.. how do you think ?

Issue:


shahednasser:
You're right this looks much better. Thank you!

PR #275, opened 2020-09-27
Added 90's Game Style Retro Buttons

anonyda:
Hi, I have added three 90's style retro buttons.

Issue: I was not able to give them a border, and I couldn't figure out why. Any help will be appreciated.
I wanted to create buttons like this: 
image
But I couldn't get a border.


shahednasser (changes requested):
Hi @anonyda thank you for your work!  

I have a few remarks and questions: 

1. Can you please add "Retro Button" under a new type called "Special Buttons"? This will align better with the "Special" section in button ideas as well. You will need to create a new file called "_special.less" in `src/components`, then create `_retro.less` in `src/components/special/` (You will need to create this folder as well).
2. After making the above change, change `buttons-examples.js` to add a new type called special, with classes as "", then add "retro-btn" inside it as a child type (You can refer to animated buttons to see how it is done).
3. I see you have three retro button types, you can add them like fill buttons (for example: Fill Button - Right). You can refer to `src/components/animated/_fill.less` for some help on how to add these button types as subtypes to `retro-btn`. 
4. Regarding the border, I tried it and was able to add the border with no trouble. For example: `border: 3px solid #333`. Please describe what you're trying to do that isn't working.

If you need any assistance on any of this, please ask!

shahednasser:
One final question: Where did you get the font? I prefer we use Google fonts. If you can find the font on Google fonts please import it. If it's not on there, please link me to look at the license of the font. Either way, we need to include a note about what font we are using for this button (You can see what I mean in Icon Buttons and the note about font awesome). It seems you didn't commit the button as well so I can't really see it on my side.

anonyda:
Hey @shahednasser,
I'll add the button under special buttons, and do the other tasks related to making a folder and adding the example. 
I'll also add the fill-btn class and see how that works out, thank you for suggestion there!
Regarding the font, it is called 'Diary of an 8-bit Mage', and I got it from FontSpace. If you want, I can even replace it with a Google Font, but I was not able to find a similar one. Anyway, just let me know.

shahednasser:
On a second thought, I don't think it's necessary to add subtypes like in fill-btn as these buttons will not have different colors for each type (Continue will just be green, reset will just be red, etc...). So just add them each under one retro type.

As for the font, can you please link me to where you got it? I need to check if we can use it and how we can inform whoever wants to use the font about how to use it.

anonyda:
Awesome,  because that was how my initial thought like... and that is why I followed the way Social Buttons were declared in the `buttons-examples.js` and used the `variation` property.

Also, here's the font link: https://www.fontspace.com/diary-of-an-8-bit-mage-font-f28455

shahednasser:
Yea I get what you mean. We still need to add it under "Special" buttons, but other than that you can do it the same way.

As for the font it looks like we can use it. We still need to add a note about it. Maybe add it at the end of "How to Use" section for now similar to font awesome, but instead of a code block just add a link to it. We will probably look for another way to add these resources later, but let's do that for now.

anonyda:
I'll do that. 
Also, I found a similar font style on Google Fonts. Here it is: https://fonts.google.com/specimen/Press+Start+2P?query=Press+

I'll make changes to the code!

shahednasser:
Great then let's use that one it will be easier. Import it in the `_retro.less` file if you can instead of directly to the website. That way if someone wants to use this button it will be ready in `sbuttons.css` without any additional imports for them.

shahednasser:
> Great then let's use that one it will be easier. Import it in the `_retro.less` file if you can instead of directly to the website. That way if someone wants to use this button it will be ready in `sbuttons.css` without any additional imports for them.

In this case, we don't need to add anything in the How to Use section.

anonyda:
Where shall I add the Google Font link?

shahednasser:
Please add it in the less file for the retro button

PR #288, opened 2020-09-28
added splash button

ub2906:
Issue:

### DIFF
--- a/src/components/animated/_splash.less
+++ b/src/components/animated/_splash.less
@@ -1,13 +1,10 @@
 .splash-btn {
-  width: 220px;
-  height: 50px;
   border: none;
   outline: none;
   color: #fff;
   cursor: pointer;
   position: relative;
   z-index: 0;
-  border-radius: 10px;
 }
 
 .splash-btn:before {
--- a/src/components/animated/_splash.less
+++ b/src/components/animated/_splash.less
@@ -1,0 +1,76 @@
+.splash-btn {
+  width: 220px;
+  height: 50px;
+  border: none;
+  outline: none;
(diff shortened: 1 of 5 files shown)


ub2906:
I have created a splash button which when clicked on, creates a moving linear colored gradient in the button

shahednasser (changes requested):
I like your idea, but can we make the animation continuous instead of on `:active`? I think it will look better that way.

ub2906:
Hi! Thanks for the quick review. I made the animation appear on hover instead of active is that ok? Or should the black buttons be replaced as well?

ub2906:
I just added a new commit as per my understanding. Hope it is better now :)

src/components/animated/_splash.less
```diff
@@ -0,0 +1,74 @@
+.splash-btn {
+  width: 220px;
+  height: 50px;
+  border: none;
+  outline: none;
+  color: #fff;
+  cursor: pointer;
+  position: relative;
+  z-index: 0;
+  border-radius: 10px;
```
shahednasser, line 10: Can we make the border radius only for the buttons that don't have class `rounded-btn`? I think this is the only change we need to make. The rest looks good

shahednasser (changes requested):

src/components/animated/_splash.less
ub2906: Done!

ub2906:
Done!

shahednasser:
Great work thank you!

@all-contributors please add @ub2906 for code and idea

PR #292, opened 2020-09-29
block button added

### LINKED ISSUE #289: Add Block Button class
Add a class that would change any button to a block button. We should also add a note about it, maybe in the "How to Use" section with an example (based on Basic Button).

This should work on any button type, so give it a basic class name something like "block-btn"

zahra-abdullah:
Issue: #289


zahra-abdullah:
Hi @shahednasser! 
I was thinking, now we have one more basic button design, how about add basic buttons, rounded buttons and block buttons in tabs, so users dont have to scroll down to see different button designs!! Let me know what you think. I can work on it as a separate fix

shahednasser (changes requested):
It looks great just one note:

1. Hinge block buttons are overlapping with the text classes under it. You can fix this by adding in `assets/css/index.css` a rule something like this:
```
#hinge .button-container .sbtn {
margin-bottom: 1.7em;
}
```
it might require `!important` so you can add it if necessary.

shahednasser:
> Hi @shahednasser!
> I was thinking, now we have one more basic button design, how about add basic buttons, rounded buttons and block buttons in tabs, so users dont have to scroll down to see different button designs!! Let me know what you think. I can work on it as a separate fix

I like this idea. Please work on it.

zahra-abdullah:
> It looks great just one note:
> 
> 1. Hinge block buttons are overlapping with the text classes under it. You can fix this by adding in `assets/css/index.css` a rule something like this:
> 
> ```
> #hinge .button-container .sbtn {
> margin-bottom: 1.7em;
> }
> ```
> 
> it might require `!important` so you can add it if necessary.

updated! thanks for the hint! Please review!

shahednasser:
Great work thank you!

PR #317, opened 2020-09-30
fix: fixed class name as per required convention and also made linting better

Subham-Panda:


shahednasser:
Thank you for your work @Subham-Panda but someone else is assigned to this issue. Please pick issues that no one is working on!

PR #318, opened 2020-09-30
footer design changes

alabo-excel:
I have improved the footer design to the best of my ability.
I hope you like what i have done.
thank you for letting me work on your project.


shahednasser:
I'm still not sure if this is the best we can do but as far as what you have done I think there's a problem with the contrast between the text and the footer. The text should be something light like white instead of black.

Issue #300, opened 2020-09-29
Header links not responsive


HADES-01:
**Describe the bug**
The header links overflow to bottom in mobile view.

**Expected behavior of what should have happened (if applicable)**
They should stay inside the header or come inside the side menu

**Screenshots (if applicable)**
![Header link bug](https://user-images.githubusercontent.com/52060608/94580230-4bcd6000-0297-11eb-93de-ecd818e91e64.png)

shahednasser:
Thank you @HADES-01 for reporting this bug. Upon testing it it seems that this occurs on devices with width <= 368px so maybe it would be best to hide those links on devices with width <= 368px.

HADES-01:
@shahednasser I can get started on it. If you'd like.

shahednasser:
sure I'll assign you @HADES-01

PR #321, opened 2020-09-30
Fixed responsiveness issues #300

HADES-01:
Fixed the position issue of header icons in screens with width <= 368px

Fixes: #300


shahednasser:
Awesome thank you!

@all-contributors please add @HADES-01 for bug

PR #328, opened 2020-09-30
add-button-idea

Nicolakacha:
I add a new button idea. It is about a play game button which has a game controller shape similar to the PS4 one.


shahednasser:
Awesome thank you!

@all-contributors please add @Nicolakacha for idea

Issue #314, opened 2020-09-30
Change font size of social-btn to 1.2rem


shahednasser:
Currently the font-size of `.social-btn` is 1rem. Change it to 1.2rem

arpancodes:
hello @shahednasser I would like to take it up for hacktoberfest, can you please assign it to me? Thanks! 😄

yuvraj23092001:
Hello I would like to contribute if you allow so .
Thanks

shahednasser:
Hi @arpancodes you can go ahead as you were first to claim it.

@yuvraj23092001 please find other issues to contribute to!

tahseenzahra:
hello, @shahednasser can I do this challenge is it available?

shahednasser:
HI, @tahseenzahra this issue has been resolved. Please look at other issues you can contribute to!

PR #331, opened 2020-09-30
change font-size of .social-btn

arpancodes:
changed font size of .social-btn from 1rem to 1.2rem

 #Issue: #314


shahednasser:
Great thank you!

@all-contributors please add @arpancodes for code

PR #347, opened 2020-09-30
added social button ideas

Neeraj3508:
added multiple ideas for social buttons.
1. neon glow button with social icon only.
2. neon glow button with social icon and name.


shahednasser:
Great thank you!

Neeraj3508:
He6 @shahednasser  i was making this new button category so do i have to make changes to buttons. Js file myself and if so what to add in variations

shahednasser:
@Neeraj3508 Yes you have to add it yourself. "variations" is only for categories like social media buttons, i.e. categories that don't follow the layout of the website of adding example for each color buttons and for basic/rounded/block buttons. If you don't need it then don't add it.

Can I ask what the category is? Do we need it as a new category or does it fit under other present categories?

Neeraj3508:
I was woeking on special category from button ideas

shahednasser:
Oh ok! I think for Special Buttons we'll add new buttons inside the "children" key in the object. If some buttons need variations then it will be added in them rather than in the "Special Buttons" category. Basically you should make it similar to "Animated Buttons"

Neeraj3508:
Ok i was trying and if there will be any issues i will text you here

PR #351, opened 2020-10-01
Added double outline buttons.

mager19:
Issue:


shahednasser:
Thank you for your work @mager19 but I'm not a big fan of how it looks. Do you think you can do something else with it to make it look better? Also some of the button colors are not correct like black instead of yellow or white instead of black, etc...

shahednasser:
Closing this issue due to inactivity and the idea not being important

PR #353, opened 2020-10-01
Glow on hover

mjmaurya:
Glow on hover buttons added


assets/js/buttons-examples.js
```diff
@@ -21,6 +21,12 @@ $(document).ready(function () {
       classes: "basic-btn",
       children: [],
     },
+    {
```
shahednasser, line 24: Please don't add it as a new type. Add it in the children of Animated in the correct alphabetical order

src/components/animated/_glow.less
```diff
@@ -0,0 +1,18 @@
+.glow-btn {
```
shahednasser, line 1: add some transition duration something like `transition: all .2s;`

shahednasser (changes requested):
thank you @mjmaurya for your work. I'm still not sure how I feel about this idea to be honest but maybe some changes can make it look better. Can you change the color of the glow based on the color of the button itself?

shahednasser:
Closing this PR due to inactivity and the contributor's work being way behind the new updates. Thank you.

PR #354, opened 2020-10-01
Added a new button idea

confusedcoder1:
Working on this new added idea which is theme-specific, along with retro and vintage style button ideas. 
(If someone is already onto them, please let me know.)


shahednasser:
I like the idea but I don't think making any button theme specific is a good thing. Can you change the naming to just "Spooky Button"?

shahednasser:
Great work thank you!

@all-contributors please add @confusedcoder1 for idea

confusedcoder1:
Hey, got it. Just changed the name. Thanks for quick response!

PR #356, opened 2020-10-01
sButtons footer design

chaitanya4vedi:
Animated gradient for footer is added. Both the timing and colour of the animation can be modified at `assets/css/index.css`.

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -390,14 +390,14 @@ button.clipboard:focus {
 }
 
 .footer {
-  color: floralwhite;
+  color: #273746;
   background: linear-gradient(-45deg, #ee7752, #e73c7e, #23a6d5, #23d5ab);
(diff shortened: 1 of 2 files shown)


shahednasser:
I like it I think it looks good! There's one change I'd make which is I think if all the text in the footer is white it would look better.

chaitanya4vedi:
> I like it I think it looks good! There's one change I'd make which is I think if all the text in the footer is white it would look better.

But then the texts and text with links are difficult to differentiate. So we can make  `text-decoration` of the text with link as underline. Then it will be more clear

shahednasser:
Yes @chaitanya4vedi add underline for links and decrease the opacity of the link on hover instead of changing to blue.

shahednasser:
Amazing work thank you!

@all-contributors please add @chaitanya4vedi for design

PR #357, opened 2020-10-01
Add whatsapp button

quangpro1610:
add new WhatsApp button

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -221,6 +221,10 @@ $(document).ready(function () {
           classes: "weibo",
           text: "Login with Weibo",
         },
+        {
+          classes: "whatsapp",
+          text: "Login with WhatsApp",
+        },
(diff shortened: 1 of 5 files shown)


shahednasser (changes requested):
I like the button but I don't think there's such thing as "Login with whatsapp" so how about changing it to something like "Chat on Whatsapp"

shahednasser:
@quangpro1610 is there a login with Bitbucket functionality? We don't want to add anything that doesn't actually exist.

quangpro1610:
> @quangpro1610 is there a login with Bitbucket functionality? We don't want to add anything that doesn't actually exist.

Have you used Bitbucket before?  It is similar to GitHub
Read more: https://bitbucket.org/

shahednasser:
I know bitbucket but can you go on a different website and choose to "Login with Bitbucket"?

PR #372, opened 2020-10-01
play store add apple store button added

dsnehasish74:
Screenshot (285)
Screenshot (286)

Issue: https://github.com/shahednasser/sbuttons/issues/355

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -145,7 +145,12 @@ $(document).ready(function () {
           classes: "add-to-cart-btn",
           text: "Add To Cart",
         },
-
+        {
+          type: "apple-store",
+          label: "Apple Store Download Button",
+          classes: "applestore",
+          text: "Download",
+        },
         {
           type: "buy-me-coffee",
           label: "Buy Me Coffee Button",
@@ -181,6 +186,12 @@ $(document).ready(function () {
           label: "Plus Button",
           classes: "plus-btn",
           text: "",
+        },
+        {
+          type: "play-store",
+          label: "Play Store Download Button",
+          classes: "playstore",
+          text: "Download",
         },
       ],
     },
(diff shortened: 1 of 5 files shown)


assets/js/buttons-examples.js
```diff
@@ -145,7 +145,12 @@ $(document).ready(function () {
           classes: "add-to-cart-btn",
           text: "Add To Cart",
         },
-
+        {
+          type: "apple-store",
+          label: "Apple Store Download Button",
+          classes: "applestore",
+          text: "Download",
```
shahednasser, line 152: Change the label to "App Download Button" and add a new key `variations` where you add inside it the apple store and the play store buttons. You can follow the implementation of social media buttons for this.

assets/js/buttons-examples.js
```diff
@@ -182,6 +187,12 @@ $(document).ready(function () {
           classes: "plus-btn",
           text: "",
         },
+        {
```
shahednasser, line 190: remove it from here and add it above in the "App Download Button"

shahednasser (changes requested):
Thank you for your work @dsnehasish74 but we need them to be similar to the black buttons you see for downloading apps. Instead of all these different buttons with all the colors it will be just two buttons, one for ios and one for android. 

The implementation for this in the JS file is similar to social buttons. I have included in the review below instructions on how you can do it.

dsnehasish74:
ok I am doing it

dsnehasish74:
Screenshot (288)

are you asking for this type  button

shahednasser:
Yes that's the one

dsnehasish74:
can help me that how can I add this 2 layer text in the button ............ I mean in the upper example "Download on the " text part

shahednasser:
> can help me that how can I add this 2 layer text in the button ............ I mean in the upper example "Download on the " text part

My suggestion is that you use ":before" for the upper text and the lower one would just be the text of the button or you can use ":after" for that as well.

PR #383, opened 2020-10-01
Color teal & Button Black Hover

LuvRedRose:
Issue:


shahednasser:
Hi @LuvRedRose thank you for your contribution but we're currently not looking into adding any new colors, as we already have many, and adding a new color has proved a hassle before unless it's necessary.

Please contribute to any of the unassigned issues or issues with the label "Always Open".

PR #390, opened 2020-10-01
Formatted texts

Pranav108:
Previous text formatting was not easy to read now it is improved.

Issue: Text formatting is Improved, previous was not easy to read, not it's better.


Pranav108:
Improved text formatting

shahednasser:
Thank you for your contribution @Pranav108 but the file is minified so that's why it looks like that. Also, we have removed that file long ago so make sure your repository is up to date.

PR #394, opened 2020-10-01
Added Circular Tick Button

akshatgarg12:
Issue:


shahednasser:
We can't add an "on click" event as this is a CSS only project. How about you change the idea to just a CSS Loading button where it's always loading.

PR #459, opened 2020-10-01
Added retro.less to 'special' buttons folder

anonyda:
I have added `_retro.less` as a special button type, along with the special button directory. 
Also added Google Font, and made changes in the button-example.js file.

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -235,6 +235,25 @@ $(document).ready(function () {
       ],
       children: [],
     },
+    {
+      type: "special",
+      label: "Special Buttons",
+      classes: "",
+      variations: [
+        {
+          classes: "retro-btn",
+          text: "Continue",
+        },
+        {
+          classes: "retro-btn-proceed",
+          text: "Proceed",
+        },
+        {
+          classes: "retro-btn-reset",
+          text: "Reset",
+        },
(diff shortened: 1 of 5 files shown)


shahednasser (changes requested):
I really like what you've done it looks great! There's only a few changes I would like you to make but not in the design, just for the layout of the website:

In buttons-examples.js for special, remove the variations array and add a children array (similar to animated), and then add an object in the array for the retro having key "variations" with the variations. For example:

```
{
      type: "special",
      label: "Special Buttons",
      classes: "",
      children: [
      {
           type: "retro",
          label: "Retro Buttons",
          variations: //The variations you added previously
       }
      ]
}
```

Please let me know if you need any help with this or if you run into any problem.

anonyda:
Hey, wouldn't that lead the buttons to inherit the colors and other styles?

shahednasser:
> Hey, wouldn't that lead the buttons to inherit the colors and other styles?

No because the script in `buttons-examples.js` detects the variations key (even in children buttons) and executes it differently. If you run into any problems though please let me know.

anonyda:
Alright, I did it!
Shall I open another PR?

shahednasser:
Great work! Thank you so much!

@all-contributors please add @anonyda for code and design

PR #464, opened 2020-10-01
scroll to top icon

adsingh14:
Added Scroll-to-Top icon without text.

Issue:

### DIFF
--- a/src/components/icon/_scroll-to-top.less
+++ b/src/components/icon/_scroll-to-top.less
@@ -15,8 +15,8 @@
   }
 
   &.rounded-btn {
-    width: 2em;
-    height: 2em;
-    padding: 0 0 0 3px;
+    width: 1.5em;
+    height: 1.5em;
+    padding: 0 0 0 1px;
   }
 }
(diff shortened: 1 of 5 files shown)


src/components/icon/_scroll-to-top.less
```diff
@@ -0,0 +1,22 @@
...
+    text-rendering: auto;
+    -webkit-font-smoothing: antialiased;
+  }
+
+  &:not(.rounded-btn) {
+    padding: 0 0.5em;
+  }
+
+  &.rounded-btn {
+    width: 2em;
```
shahednasser, line 18: change width and height to 1.5em and padding to `0 0 0 1px`

shahednasser (changes requested):
I just have one change to what you've done.

src/components/icon/_scroll-to-top.less
adsingh14: Fixed !

shahednasser:
Great thank you!

@all-contributors please add @adsingh14 for code

Issue #444, opened 2020-10-01
Add action to comment on PR after deploying it


shahednasser:
We have an action called [Surge PR Preview](https://github.com/marketplace/actions/surge-pr-preview) that previews a PR on a surge website. This is very helpful as it makes it easier to see the changes of the PR.

Currently, the action does not leave a comment with the URL of the deployed PR, it just deploys it but it appears in the action log. This is apparently not possible due to github limitation on adding a comment from an action on a PR.

I'm not sure if it's possible but I would like to add an action that submits a comment on the PR after the Surge PR Preview action succeeds with the URL of the deployed PR. This will make it easier for me and the contributors as well to see their changes.

If it's not possible with this action and you know of a better action to do this please let me know as well. (It doesn't have to be surge but it's the easiest).

HeinKhantZaw:
Can I work for this issue?

shahednasser:
Sure @HeinKhantZaw. Let me know how you want to do it.

HeinKhantZaw:
Hi @shahednasser . I don't know if it's going to work or not but I am thinking about using Nelify deploy action. How about that?

HeinKhantZaw:
![image](https://user-images.githubusercontent.com/40730345/94842146-0c427780-0441-11eb-9d24-74d2572b5219.png)

Here is the action for comment to PR I've tested in my repo. It's used by netlify deploy action.

shahednasser:
Is netlify free? Also we need to confirm that it works on forked repositories so we will need to test it. I'll let you know with more details in a bit.

HeinKhantZaw:
Yes. It is free for personal projects. I don't know exactly if it works for forked repos but you can check more here.
[https://docs.netlify.com/#get-started](url)

HeinKhantZaw:
I will submit pull request for preview.yml update if everything is ok. Please let me know again.

shahednasser:
I have created a new branch called preview-action. Please submit a new PR to merge the changes to it so we can test it on that branch.

shahednasser:
Are there any keys I'm required to add to this repo? Please let me know as well.

HeinKhantZaw:
You need to add NETLIFY_AUTH_TOKEN and NETLIFY_SITE_ID.
For NETLIFY_AUTH_TOKEN, you can get it from here [personal access token](https://app.netlify.com/user/applications#personal-access-tokens)
[...]

PR #479, opened 2020-10-01
Update preview.yml for comment action

HeinKhantZaw:
Issue: Fix for #444

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,20 +1,51 @@
-name: Surge PR Preview
+name: PR Preview
 
 on:
+  push:
   pull_request:
-    paths-ignore:
-      - '**.md'
-      - '.all-contributorsrc'
+    types:
+      - opened
+      - synchronize
 
 jobs:
-  preview:
+
+  # This job will:
+  # -deploy a draft every time there is a pull request created or synchronized that is not on master branch
+  # -comment on that pull request with the deploy URL
+  deployPRDraft:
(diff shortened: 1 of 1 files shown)


.github/workflows/preview.yml
```diff
@@ -1,20 +1,51 @@
...
+          netlify-auth-token: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+          netlify-site-id: ${{ secrets.NETLIFY_SITE_ID }}
+          build-dir: '.'
+          draft: true
+          comment-on-pull-request: true
+
+  # This job will:
+  # -deploy a production build every time there is a push on the master branch
+  # -comment that commit with the deploy URL
+  publishMasterCommit:
```
shahednasser, line 36: Is this on pushing to the master branch? We don't need an action for that as we are using github pages for the moment, but please comment it out in case we need it later on.

shahednasser (changes requested):
I have also added the keys you requested in the issue

.github/workflows/preview.yml
HeinKhantZaw: Ok. So, do I need to change the action for every PR to master branch too?

shahednasser: We just need the action to run when a PR is submitted. It shouldn't run on merge or push to master

shahednasser:
Ok next we need to test it out. Can you try creating another pull request to preview-action branch so we can see if it's working? It can be any change this is just for testing.

PR #481, opened 2020-10-01
keyBtn

Charu271:
Issue:

### DIFF
--- a/src/components/animated/_key.less
+++ b/src/components/animated/_key.less
@@ -1,0 +1,106 @@
+.key-btn.pink-btn {
+  border-top: 2px solid @pink;
+  background-color: @pink;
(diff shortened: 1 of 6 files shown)


src/components/animated/_key.less
```diff
@@ -0,0 +1,106 @@
...
+  display: inline-block;
+  background-color: transparent;
+  font-size: 1.5rem;
+  color: white;
+  font-weight: bolder;
+  text-decoration: none;
+  padding: 15px 25px;
+  border-radius: 4px;
+  position: relative;
+  display: block;
```
shahednasser, line 97: remove display: block

shahednasser (changes requested):
Looks good I just have one change. Don't forget to run `npm run compile` afterwards.

PR #488, opened 2020-10-01
testing for PR action

HeinKhantZaw:


HeinKhantZaw:
@shahednasser Did u add netlify-auth-token in your repo? The error shows that "Input required and not supplied: netlify-auth-token"

shahednasser:
Yes I added NETLIFY_AUTH_TOKEN and NETLIFY_SITE_ID. I'm trying to figure out what the issue is too

shahednasser:
The issue is that it cannot access Github Secrets from a forked repository. It says so in the Secrets page:

"Secrets are not passed to workflows that are triggered by a pull request from a fork."

We'll need to look to a better way to implement this I guess.

shahednasser:
@HeinKhantZaw can you look into [this](https://github.com/marketplace/actions/netlify-actions)? Do you think it will work better?

shahednasser:
@HeinKhantZaw Can you please create another test PR just to make sure it's all working correctly?

PR #490, opened 2020-10-01
Special diagonal rounded button.

adsingh14:
Added: Special Button with diagonally-curved corners.

Issue: For this button, there's should not included `.rounded-btn`, it's already curved from corners.


src/components/special/_rounded-diagonal.less
```diff
@@ -0,0 +1,19 @@
+.rounded-diagonal-btn {
+  background: #7c26c2;
```
shahednasser, line 2: Instead of giving it a background color, use one of the colors we have in the example (for example purple-btn). This will help the person using this button to change the color by using whatever colored classes we have.

src/components/special/_rounded-diagonal.less
```diff
@@ -0,0 +1,19 @@
...
+  font: bold 24px "Syne";
```
shahednasser, line 3: remove font-family "Syne" as this is set in the website but shouldn't be set in sbuttons.css

shahednasser (changes requested):
Thank you for your contribution. It looks good but I have a few comments below. Additionally, from a ux perspective, the button when clicked should give feedback to the user that they clicked it (For example, adding box-shadow on active state). Please add something similar that you find fitting for this button.

shahednasser:
Also there seems to be some problems with other buttons after you merged the master to your branch. Please fix this.

src/components/special/_rounded-diagonal.less
adsingh14: I've set that font to overlap to cursive-font which doesn't match with the button theme. And 'retro button' .less file is also having the custom font.

shahednasser: In retro button the font is imported in the Less file and it is necessary to the button's design. Syne is not imported in the Less file it's just the font of the website. If you want to change the font-family to Syne then set it in `assets/css/index.css` as it is not required for the button itself.

src/components/special/_rounded-diagonal.less
adsingh14: I think that's special category button which is why I used css file for background-color.

adsingh14: It lies under special category, instead of using 'purple-btn' from js, I prefer to use the same color from **sbuttons.less**.

adsingh14: If I add second class from js file then it won't show the period in `class snippet`.

image

adsingh14: Added !

adsingh14:
Recent update button shadow is working with both theme.

shahednasser:
It looks great but I see now that the added button's "rounded-btn" section is unnecessary. Maybe we can add a key in the javascript option, something like "shouldHaveRounded" and then in `createSection` it checks if this key is in the object and it is set to false then we shouldn't add a rounded buttons section. We'll also have to add it in the documentation in CONTRIBUTING.md. Can you do that?

adsingh14:
> 
> 
> It looks great but I see now that the added button's "rounded-btn" section is unnecessary. Maybe we can add a key in the javascript option, something like "shouldHaveRounded" and then in `createSection` it checks if this key is in the object and it is set to false then we shouldn't add a rounded buttons section. We'll also have to add it in the documentation in CONTRIBUTING.md. Can you do that?

I was referring about that issue. I can do it, but need some time.

shahednasser:
@adsingh14 no problem take your time!

adsingh14:
I was thinking to add that **key** but my focus was on button.

shahednasser:
If you would like to add it please add it to this PR as it will make the button look better in terms of presentation.

But if you don't want to work on it that's fine too! Let me know.

adsingh14:
I can do it. Figuring out the examples.js workflow.

adsingh14:
I have added key for 'Rounded button' class. If anyone use `shouldHaveRoundedType: false`, then it will hide '.rounded-btn' section.

shahednasser:
Looks great thank you for your work!

@adsingh14 we have created a new organization for this repo and added a team of reviewers. I think you have been able to grasp the idea of the project fast and have been a great help! Would you be like to added as a reviewer in this repo?

shahednasser:
@all-contributors please add @adsingh14 for idea

adsingh14:
> 
> 
> Looks great thank you for your work!
> 
> @adsingh14 we have created a new organization for this repo and added a team of reviewers. I think you have been able to grasp the idea of the project fast and have been a great help! Would you be like to added as a reviewer in this repo?

Reviewer 🤔... Okay, let me in ! I would love to provide ideas and be a part of this project.

PR #491, opened 2020-10-01
Searched for rules in index.css only used by getStarted.html, moved t…

baspalinckx:
Searched for rules in index.css only used by getStarted.html
Moved these rules to a new file: getStarted.css
Added an import for this very file in getStarted.html

Issue: https://github.com/shahednasser/sbuttons/issues/485

### DIFF
--- a/assets/css/getStarted.css
+++ b/assets/css/getStarted.css
@@ -1,0 +1,54 @@
+.section-header {
+  text-align: center;
+}
+
+.secondary-hr {
+  width: 10%;
+  border-color: rgba(27, 114, 145, 0.3);
+}
+
+.text {
+  font-size: 1.3rem;
+  color: black;
+  word-wrap: break-word;
+  font-family: "Syne", sans-serif;
+}
+
+.code-text {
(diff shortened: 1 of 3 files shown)


assets/css/getStarted.css
```diff
@@ -0,0 +1,54 @@
+.section-header {
```
shahednasser, line 1: this is used by index.html so it needs to be in index.css

assets/css/getStarted.css
```diff
@@ -0,0 +1,54 @@
...
+  text-align: center;
+}
+
+.secondary-hr {
+  width: 10%;
```
shahednasser, line 6: same for this one

shahednasser (changes requested):

baspalinckx:
Really, I could not find both of these with a CTRL-F search in index.html? But I can imagine that they might be used there in the future. Changed it

shahednasser:
Great work thank you!

@all-contributors please add @baspalinckx for code

PR #493, opened 2020-10-01
Preview action update

HeinKhantZaw:

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,6 +1,5 @@
 name: Build and Deploy to Netlify
 on:
-  push:
   pull_request:
     types: [opened, synchronize]
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/preview.yml
```diff
@@ -1,5 +1,6 @@
 name: Build and Deploy to Netlify
 on:
+  push:
```
shahednasser, line 3: Please remove or comment out push as we don't need it.

HeinKhantZaw:
@shahednasser I guess It's working now
IMG_20201002_040354

.github/workflows/preview.yml
HeinKhantZaw: ohh..sry about that, I was working with my phone in my bed as it's 4 a.m right now.

shahednasser:
Great thank you!

@all-contributors please add @HeinKhantZaw for code and tool

Issue #309, opened 2020-09-30
Improve style of `.social-btn` with class `block-btn`


shahednasser:
Social Buttons that are also blocks have the icons at the very far right. It should be next to the "Login with ..." text

chaitanya4vedi:
@shahednasser Can i do this?

shahednasser:
Yes @chaitanya4vedi I'll assign you

pinkman7009:
can i work on this?

shahednasser:
@pinkman7009  Someone is already working on this. please choose an unassigned issue

chaitanya4vedi:
> Yes @chaitanya4vedi I'll assign you

@shahednasser you can unassign me as I might take some more time to do this and somone is already interested in this

pinkman7009:
@shahednasser so can i work on it now?

shahednasser:
Sure @chaitanya4vedi .

Assigning @pinkman7009

pinkman7009:
thank you. i will start working on this now.

adsingh14:
Social buttons not vertically centered align. I can fix this.

adsingh14:
Social buttons not centered aligned vertically . I can fix this.

tushar2407:
Hi.! The icons are not aligned with respect to text in Y-axis. I can fix that. Can you assign me this?

shahednasser:
@adsingh14 @tushar2407 someone is assigned to this issue. Please look for unassigned issues.

PR #500, opened 2020-10-01
social block buttons moved towards login text

pinkman7009:
The icons of the social buttons which are also block buttons have been moved a bit left towards the "Login with.." text.

Files changed : dist/sbuttons.css and src/components/_social.less
Issue: #309


shahednasser:
Looks better thank you!

@all-contributors please add @pinkman7009 for code and bug

PR #502, opened 2020-10-01
Feature/pudding button

baspalinckx:
Added a button which will jiggle on hover, like a pudding!
Only issue is for the life of me, I can't get it to show up in sbutton.css. I do exactly as is described on the page. What am I missing? 

Issue: https://github.com/shahednasser/sbuttons/issues/498


src/components/animated/_pudding.less
```diff
@@ -0,0 +1,35 @@
+.pudding-btn() {
```
shahednasser, line 1: The problem is here. When you add `()` in less it creates a mixin not a class. Remove them to get it to work

shahednasser (changes requested):
Make sure to compile after doing the change below

src/components/animated/_pudding.less
baspalinckx: Of couse! How stupid of me

baspalinckx:
Should be good now!

assets/js/buttons-examples.js
```diff
@@ -126,6 +126,11 @@ $(document).ready(function () {
           label: "Splash",
           classes: "splash-btn",
         },
+        {
```
shahednasser, line 129: Looks cute but my only remark is to place this button in the correct alphabetical order under "Animated" (sorry I didn't notice this before).

baspalinckx: No problem. Should have spotted that

shahednasser:
Great work thank you!

@all-contributors please add @baspalinckx for idea

Issue #355, opened 2020-10-01
Playstore and apple store download button want to add .


dsnehasish74:
I want to add this 2 new buttons to the site . This button idea is mentioned in the button idea file . As no one is assigned this work till now then can I plz do this ........... will you plz assign this to me !!!!! thanks

shahednasser:
Ok @dsnehasish74 I'll assign you!

dsnehasish74:
Thank you ...

PR #519, opened 2020-10-02
Google play store and apple store button added

dsnehasish74:
Screenshot (289)

Issue: #355


src/components/appstore/_playstore.less
```diff
@@ -0,0 +1,25 @@
...
+    background-color: #000000;
+    color: #fff;
+
+    &:before {
+      color: #fff;
+      content: "\f3ab";
+    }
+
+    &:after {
+      font-family: "Font Awesome 5 Brands";
```
shahednasser, line 12: Remove font family here as it is unnecessary

src/components/appstore/_applestore.less
```diff
@@ -0,0 +1,25 @@
...
+    background-color: #000000;
+    color: #fff;
+
+    &:before {
+      color: #fff;
+      content: "\f179";
+    }
+
+    &:after {
+      font-family: "Font Awesome 5 Brands";
```
shahednasser, line 12: Remove font family here as it is unnecessary

shahednasser (changes requested):
Looks great I just need you to do a few changes for me:

1. Move the button into the Icon Buttons type. 
2. Instead of creating `src/components/_appstore.less` create it in `src/components/icon/_appstore.less` and move the styling in `_applestore.less` and `_playstore.less` to it.

dsnehasish74:
Ok .. but one question If I add it there in the icon section then I think it will add all the the color button which we don't need???

src/components/_appstore.less
```diff
@@ -0,0 +1,35 @@
+.appstore-btn {
+  box-shadow: 0px 1px 4px #4d4d4d;
+  transition: all 0.2s;
+  position: relative;
+  padding-top: 15px;
+  padding-left: 50px;
+  font-size: 1.5rem !important;
+  font-family: "Product Sans" !important;
```
shahednasser, line 8: Also remove this font-family

shahednasser:
@dsnehasish74 no if you put the exact same object you made inside icon buttons it won't.

Can you also add in `assets/css/index.css` a rule to use "Syne" font for this button?

assets/js/buttons-examples.js
```diff
@@ -128,6 +128,22 @@ $(document).ready(function () {
...
+        {
+          classes: "applestore",
+          text: "App Store",
+        },
+        {
+          classes: "playstore",
+          text: "Google Play",
+        },
+      ],
+      children: [],
```
shahednasser, line 145: Remove this empty array

dsnehasish74:
I add this under icon buttons ............... can you check this now

shahednasser:
It looks better but you still need to do two things indicated above: 

1. move the styling in `_applestore.less` and `_playstore.less` to `src/components/icon/_appstore.less`
2. add in `assets/css/index.css` a rule to use "Syne" font for this button

dsnehasish74:
As the upper texts are different in two badges so if I shift the style from _applestore.less to _appstore.less then How can  I maintain this two different texts ?????????????????????? 
2 .I can not understand this part ??????? rule mean should I add this in that file or you are talking about something else
.appstore{
fontfamily: "syene";
}

shahednasser:
As for the first part, you can still maintain it since they have different class names. I'm not sure what the issue exactly.

As for the second part, that's what I mean but the font is called "Syne" so you just need to add that in `assets/css/index.css`

shahednasser:
Hi @dsnehasish74 I think you misunderstood. I meant to move the styles in `_playstore.less` and `_applestore.less` to `_appstore.less` not the opposite.

dsnehasish74:
That mean I have to delete this two files? and move all the styles in this _appstore.less file??? is it???

shahednasser:
> That mean I have to delete this two files? and move all the styles in this _appstore.less file??? is it???

Yes

dsnehasish74:
Is it okay now?

src/components/icon/_appstore.less
```diff
@@ -0,0 +1,78 @@
...
+
+    &:before {
+      color: #fff;
+      content: "\f3ab";
+    }
+
+    &:after {
+      position: absolute;
+      content: "Get it on";
+      top: 0px;
```
shahednasser, line 68: change to 3px

src/components/icon/_appstore.less
```diff
@@ -0,0 +1,78 @@
...
+
+    &:before {
+      color: #fff;
+      content: "\f179";
+    }
+
+    &:after {
+      position: absolute;
+      content: "Download on the";
+      top: 0px;
```
shahednasser, line 45: change to 3px

shahednasser (changes requested):

shahednasser (changes requested):
It looks great but there's one last thing I'm noticing. Can you add something specific for playstore block buttons? I think the left needs fixing. Set the left to 87px

src/components/icon/_appstore.less
```diff
@@ -66,7 +66,7 @@
       position: absolute;
       content: "Get it on";
       top: 3px;
-      left: 100px;
+      left: 87px;
```
shahednasser, line 69: Don't change it for all buttons just change it for buttons that have the class `.block-btn`

dsnehasish74: yes but this change will only reflect on block btn because for non block button style is left: 50px  ; so indirectly it is only for block btns ... I guess

shahednasser:
Alright looks great thank you!

@all-contributors please add @dsnehasish74 for code

dsnehasish74:
Thank u very much for helping me!!!! :)

shahednasser:
Thank you for being patient in resolving the comments and issues! I appreciate your contribution.

adsingh14:
**Alignment in block buttons.**

image
image

dsnehasish74:
What device you are using?????????????????? because in my pc it is ok for any kind of screen size , Don't know why it is showing this type of error in your device🤔

shahednasser:
It looks okay on mine as well. @adsingh14 you can submit an issue with more details on the device, browser, screen width etc...

adsingh14:
@dsnehasish74 I'm using 15:6 screen size laptop. Resolution : 1366 x 768 (in px).
Tested it on latest version Mozilla Firefox, Chrome.

That's a bug not an error.

shahednasser:
Please send an issue with it instead of commenting here.

PR #558, opened 2020-10-02
added glow button

malihanan:
Issue:


src/components/animated/_glow.less
```diff
@@ -0,0 +1,95 @@
+:root {
+  --light: 80;
```
shahednasser, line 2: Instead of adding these variables on root you should add them as less variables just like other variables in the files (You can refer to `src/sbuttons.less` for this

shahednasser (changes requested):
Great work! I just have one comment below about the code.

src/components/animated/_glow.less
malihanan: Sure, will do that

malihanan: Should I add them in src/sbuttons.less itself?

shahednasser: Yes add them there

shahednasser: We will be moving all variables to a different file later on

malihanan: Okay

malihanan: Did that, can you check and confirm?

src/components/animated/_glow.less
```diff
@@ -1,12 +1,7 @@
-:root {
-  --light: 80;
-  --threshold: 60;
-}
-
 .glowBorderRule(@color) {
   border-color: @color;
   background-color: rgba(@color, 0.2);
-  --switch: calc((var(--light) - var(--threshold)) * -100%);
+  --switch: calc((var(light) - var(threshold)) * -100%);
```
shahednasser, line 4: Change var(light) to just @light. same for threshold. Also add switch to the variables like you did before.

malihanan: I realized those variables don't make sense. I was trying to change font color based on background but realized that is on user to modify based on their background as the button itself is translucent. Sorry for the confusion, it's my first time using less. Can you check now?

shahednasser:
Looks great thank you!

@all-contributors please add @malihanan for code and idea

shahednasser:
@all-contributors please add @malihanan for code and idea

malihanan:
> Looks great thank you!
> 
> @all-contributors please add @malihanan for code and idea

Was great contributing!

Issue #480, opened 2020-10-01
Add comments in JS files


adsingh14:
**Is your feature request related to a problem? Please describe.**
Code is not formatted. 

**Describe the solution you'd like**
Single/block line comment indeed ! It would be great to add comments before committing code.  
`// Single line code for rare but useful checkpoints `
`/* Block lines comment
    for well documented code */`

**Additional notes**
I know the minified files won't include any comment but it will be useful for devs.

shahednasser:
Do you mean comments in the less files?

adsingh14:
Extremely needed in JS and css ((un-minified versions).

palakyadav1807:
Can you assign me this for .html file?

shahednasser:
@adsingh14 I don't think it's necessary in the CSS but you're right about the JS. I'll change the title to make it more fitting. Let me know if you would like to work on it.

HADES-01:
@shahednasser Should the comments be detailed or just general explanation of functions? I can get on it either way.

shahednasser:
@HADES-01 I guess it should be enough to make it understandable for contributors. I'll assign you.

adsingh14:
I will add single line comments in main **.js** file.

shahednasser:
Ok @HADES-01 you can work on `assets/js/buttons-examples.js` and @adsingh14 you can work on `assets/js/index.js`

adsingh14:
Sounds good ! Later on, I will follow JSDoc guidelines for block comments and I recommend the same to other assignee.

HADES-01:
@adsingh14 Yeah sure.

shahednasser:
@HADES-01 added good documentation in buttons-examples.js but now I'm thinking maybe we should also include in the documentation the keys allowed in `buttons` array as it can get confusing. What do you guys think?

HADES-01:
@shahednasser Yeah, I actually was going to create a new issue regarding this very feature. We should include it in the contributors guidelines so that it removes all the ambiguity for the contributors.

shahednasser:
@HADES-01 it's already in the contributing guidelines but I think we should add it in the jsDoc as well. Let me know if you can do that

HADES-01:
@shahednasser Just to be sure you want me to add `jsDoc` for the keys of the button object in the `buttons-examples.js`.

shahednasser:
@HADES-01 yes

HADES-01:
[...]

PR #576, opened 2020-10-02
Added JSDoc for the buttons list objects

HADES-01:
Added JSDoc documentation to buttons list object

Issue: #480

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -16,14 +16,15 @@ $(document).ready(function () {
   /**buttons is a list of button type objects tht will be rendered on the page load
   * with their respective class names, labels, children and types.
   * 
+  * properties with [] are optional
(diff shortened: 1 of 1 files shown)


shahednasser (changes requested):
Looks great but can you also add which properties are optional and what the default would be? Everything but type and label is optional.

HADES-01:
@shahednasser Isn't classes a required property ??

shahednasser:
@HADES-01 no it's not. For example, Animated can work without the classes property.

HADES-01:
@shahednasser Actually according to the jsDocs conventions properties with `[ ]` are considered optional but I can write it out as well if you'd like and what should be default for classes, variations and children

shahednasser:
Yes I think it's best to include it as some people may not know that. 

These properties at the moment don't have a default value so don't add it for them.

shahednasser:
Amazing thanks!

PR #588, opened 2020-10-02
Fixed overlapping navbar on sidebar

Goodiec:
Issue:


PhilippeBarre:
If I may say so, I find my z-index solution better because with your solution, the cross to close the bar is hidden.

Goodiec:
> If I may say so, I find my z-index solution better because with your solution, the cross to close the bar is hidden.

OK, let me try to fix that, thank you

PR #589, opened 2020-10-02
Fix overlapping navbar on sidebar

Goodiec:
Issue:


shahednasser:
Instead of adding padding, can you just change the z-index of the sidebar to "9999"

Goodiec:
> Instead of adding padding, can you just change the z-index of the sidebar to "9999"

@shahednasser I tried that and the navbar was still overlapping the sidebar

Goodiec:
> > Instead of adding padding, can you just change the z-index of the sidebar to "9999"
> 
> @shahednasser I tried that and the navbar was still overlapping the sidebar

I'll reduce the value of the navbar's z-index on mobile and increase that of the sidebar and see if it works.

shahednasser:
@Goodiec are you sure? I tried to just add z-index and it worked fine without any other changes. If lowering the navbar's z index helps without disrupting anything else then you may do that.

PR #598, opened 2020-10-03
Rounded top

Olsyx:
I added a new special button called rounded top.
The new class is _rounded-top.less, and I added it to buttons-examples.js and _special.less.

Issue: Add New Buttons


shahednasser:
It looks great thank you but next time please submit the idea in a new issue before working on it.

@all-contributors please add @Olsyx for code and idea

PR #599, opened 2020-10-03
Rounded Bottom

Olsyx:
Created new special button: Rounded Bottom
The new class is _rounded-bottom.less, and I added it to buttons-examples.js and _special.less.

Issue: Add New Buttons


shahednasser:
Great thanks!

Issue #433, opened 2020-10-01
[BUTTON IDEA]: Vintage Button


shahednasser:
**1. Name of button**: Vintage Button

**2. Description**: A button that has a Windows 95 Vintage style to it

**3. Button Type (Animated, Special, etc...)**: Special

**4. Will you work on it?**: no

**submitted by**: @anonyda

anonyda:
Hi @shahednasser, I am working on it

shahednasser:
Great @anonyda I'll assign you

PR #614, opened 2020-10-03
Added Windows 95 style buttons in the Specials

anonyda:
Hey, I created another button type, which closes Issue #433 
These are Windows 95 style vintage buttons.
I am using a font from Google Fonts.
Do let me know how you think of it!


src/components/special/_win95.less
```diff
@@ -0,0 +1,30 @@
...
+  box-shadow: 0 0 0 black;
+  transition: all 0.2s;
+}
+
+.win95-btn:last-child {
+  margin: 0;
+}
+
+.win95-btn:hover {
+  box-shadow: 0.4rem 0.4rem 0 black;
```
shahednasser, line 23: Reduce translate and box-shadow on hover I think it is a little too much

shahednasser (changes requested):
Is using this google font necessary? It doesn't really make it look more like a windows 95 button. Either choose a different font or don't add one.

Also in the less file can you please use less syntax? i.e, instead of repeating `.win95-btn` over and over again, place everything related to it inside it once and refer to the button with "&". Please refer to Less on more info on this.

shahednasser:
I suggest using this font instead: https://fonts.google.com/specimen/VT323?category=Monospace

src/components/special/_win95.less
anonyda: thanks for the feedback! I made the changes.

shahednasser:
Looks good thank you!

Issue #426, opened 2020-10-01
[BUTTON IDEA]: Share Button


shahednasser:
**1. Name of button**: Share Button

**2. Description**: Button with the share icon only

**3. Button Type (Animated, Special, etc...)**: Icon

**4. Will you work on it?**: no

LuciKritZ:
I would like to work on it.

shahednasser:
Sure @LuciKritZ

MeerHamza1421:
Can i work on it for hacktober?? and can i add a social media share button

shahednasser:
Someone is assigned to this.

LuciKritZ:
Hi @shahednasser ! I am actually a bit confused about choosing the icon for the share button.
I have the following choices:
1. 
![image](https://user-images.githubusercontent.com/26204432/94991180-f6e35f80-059e-11eb-83c2-dc49300eeb2f.png)
2. 
![image](https://user-images.githubusercontent.com/26204432/94991191-0b275c80-059f-11eb-8e3e-6f7361ed9b2e.png)
3.
![image](https://user-images.githubusercontent.com/26204432/94991202-21cdb380-059f-11eb-9ce8-3c10f6143176.png)
4. 
![image](https://user-images.githubusercontent.com/26204432/94991219-38740a80-059f-11eb-943a-3566f9f1c3dc.png)
5. 
![image](https://user-images.githubusercontent.com/26204432/94991227-4aee4400-059f-11eb-85c8-eb8c243ea42f.png)

For now, I have added this icon and share icon looks something like this in the main page:
![image](https://user-images.githubusercontent.com/26204432/94991250-6fe2b700-059f-11eb-835e-5a88fcac659c.png)

Do you have any suggestions? Or should I raise the PR if the icon seems okay?

shahednasser:
I prefer the 4th one

maityankan10:
Can i contribute for hacktoberfest

shahednasser:
@maityankan10 someone is working on this

PR #616, opened 2020-10-03
Share icon

LuciKritZ:
Issue: #426 

- **Name of the button**: Share Button
- **Description**: Button with the share icon only
- **Button Type (Animated, Special, etc...)**: Icon

**Screenshots**:
1. On Active (Changed from black color to neonBlack):
image

2. Normal View (Black color):
image


assets/js/buttons-examples.js
```diff
@@ -281,6 +281,20 @@ $(document).ready(function () {
             },
           ],
         },
+        {
+          type: "share",
+          label: "Share Button",
```
shahednasser, line 286: Remove "Button" from label"

src/components/icon/_share.less
```diff
@@ -0,0 +1,18 @@
...
+    font-family: "Font Awesome 5 Free";
+    display: inline-block;
+    font-style: normal;
+    font-variant: normal;
+    text-rendering: auto;
+    -webkit-font-smoothing: antialiased;
+  }
+
+  &:active {
+    color: @neonBlack;
```
shahednasser, line 16: This is not visible. Please try to show off active state in a better way.

shahednasser (changes requested):

src/components/icon/_share.less
LuciKritZ: image

Does this look good?

shahednasser: How about giving it more color? Blue or something like that

LuciKritZ: Yes, sure!

LuciKritZ: @shahednasser So, I am thinking to go with this:
image

Any suggestions?

assets/js/buttons-examples.js
```diff
@@ -287,6 +287,20 @@ $(document).ready(function () {
             },
           ],
         },
+        {
```
shahednasser, line 290: Place the button in the correct alphabetical order.

shahednasser (changes requested):

src/components/icon/_share.less
shahednasser: I think the blue looks fine but can you make a new variable called "shareActive" and set it to blue? This will make it easier to change it later on if anyone needs to

Issue #395, opened 2020-10-01
Move variables in src/sbuttons.less to new file src/_variables.less


shahednasser:
Just move the variables in `src/sbuttons.less` to new file `src/_variables.less` then import `src/_variables.less` in the beginning of sbuttons.less before any file

You should also change the link in the README for sbuttons.less under [Modifying Button Colors](https://github.com/shahednasser/sbuttons#modifying-button-colors) to the new variables.less file and change the text as well to "_variables.less"

annettomcy:
Would like to work on this

shahednasser:
Ok @annettomcy I will assign you

pinkman7009:
is anyone working on this?

annettomcy:
Yes me

pinkman7009:
okay

faraz16iqbal:
Is this issue still open?

shahednasser:
@faraz16iqbal this issue is assigned. Please look for unassigned issues.

abhishekrawe:
I want to work on it? is this still open?

shahednasser:
@abhishekrawe someone is working on this.

shahednasser:
Unassigning @annettomcy as they are taking too long and this is a high priority issue.

tmott13:
I would like to be assigned to this, if that's possible? @shahednasser

kuldeep123kd:
Can i work on this?

shahednasser:
@tmott13 I need someone to not take long with it can you do that?

shubh710:
Hi @shahednasser 
I can do it quickly.
Is the problem Statement same as
https://github.com/sButtons/sbuttons/issues/395#issue-712643930 ?

kuldeep123kd:
Yes @shahednasser  i can do it quickly.

shahednasser:
Assigned @kuldeep123kd

PR #617, opened 2020-10-03
Expanding Border

isonnymichael:
Add Expanding Border Button - Animation Button

Expanding Button

Issue: I'm not sure this is what you expected. Please review this


assets/js/buttons-examples.js
```diff
@@ -159,6 +159,11 @@ $(document).ready(function () {
           label: "Transparent On Hover",
           classes: "transparent-btn",
         },
+        {
```
shahednasser, line 162: Place this in the correct alphabetical order under Animated

shahednasser (changes requested):
Looks good but is it possible to store the color of the border in a variable? It would be helpful for someone who wants to use it but wants to modify the color.

assets/js/buttons-examples.js
isonnymichael: > Looks good but is it possible to store the color of the border in a variable? It would be helpful for someone who wants to use it but wants to modify the color.

I think it should be worked on another issue, because border style button are ont the only one this issue, and it will be nice for first timer maybe.

shahednasser: Hmm true. I'll create an issue after #395 is resolved. I think we need to organize all variables with good comments in one file so it can be easier to override any color.

shahednasser:
I'm testing it again now and I noticed a problem that appears in dark mode well. For some buttons that border is inside the button (meaning you can still see red (or the button's color) from ourside the border). It's not apparent in light mode but can you look into it (as we don't know how anyone might use it).

isonnymichael:
> I'm testing it again now and I noticed a problem that appears in dark mode well. For some buttons that border is inside the button (meaning you can still see red (or the button's color) from ourside the border). It's not apparent in light mode but can you look into it (as we don't know how anyone might use it).

Sorry I'm not test on dark mode yet. I will check it out

isonnymichael:
After get some diggin, I found the trick to make it better.. check it out

shahednasser:
Great thanks!

PR #621, opened 2020-10-03
Fixing side bar overlapping issue

Dheerajbana:
Issue:


shahednasser:
@Dheerajbana your repo is very outdated. Can you make it up to date or refork it then try again? It would be better this way.

Dheerajbana:
@shahednasser https://github.com/sButtons/sbuttons/pull/631 did as you suggested

PR #622, opened 2020-10-03
added liquid button

Neeraj3508:
Issue:


assets/js/buttons-examples.js
```diff
@@ -154,6 +154,11 @@ $(document).ready(function () {
           label: "Transparent On Hover",
           classes: "transparent-btn",
         },
+        {
```
shahednasser, line 157: Please place it in the correct alphabetical order.

src/components/animated/_liquid.less
```diff
@@ -0,0 +1,93 @@
+.liquid-btn {
+  position: relative;
+  z-index: -1;
```
shahednasser, line 3: Change it to 1 here

src/components/animated/_liquid.less
```diff
@@ -0,0 +1,93 @@
...
+    z-index: -1;
+  }
+
+  .liquidBtn(@color) {
+    border-color: @color;
+    color: #000 !important;
+    .bc(@color);
+
+    &:hover::before {
+      transition: 0.5s linear;
```
shahednasser, line 39: Can you place transition inside the original implementation of `::before` (not on hover) as it will make the transition cleaner when you move your mouse outside of the button

src/components/animated/_liquid.less
```diff
@@ -0,0 +1,93 @@
...
+      top: -120px;
+    }
+
+    &:hover::after {
+      transition: 0.5s linear;
```
shahednasser, line 44: Same comment regarding transition

src/components/animated/_liquid.less
```diff
@@ -0,0 +1,93 @@
...
+    .liquidBtn(@purple);
+  }
+
+  &.black-btn {
+    .liquidBtn(@black);
+    color: #fff !important;
+  }
+
+  &.white-btn {
+    .liquidBtn(@white);
```
shahednasser, line 83: Set the background of the button transparent for white so that you can actually see the animation.

shahednasser (changes requested):
Hi @Neeraj3508 I reviewed the problem you're having and the fix is simple, you just need to increase the z-index of the button itself. You put the z-index to the button at -1, so if you right click on the button and click inspect, you'll notice in dev tools it will show you `button-container` selected instead of the button itself. Because the z-index is so low, the normal hover action (without forced) is actualyy hovering on `button-container` instead of the button. Just change it to 1 and it will work.

Neeraj3508:
I tried that to it wasn't working but let's just try that again.

Neeraj3508:
I got it i was changing z-index for before and after but not the liquid-btn itself. Thank you @shahednasser

assets/js/buttons-examples.js
shahednasser: @Neeraj3508 please make this change.

src/components/animated/_liquid.less
Neeraj3508: Ok i will do these changes

Neeraj3508:
I will do above mentioned changes then push them

src/components/animated/_liquid.less
Neeraj3508: Setting this to transparent the button will look like this https://i.ibb.co/v4WMjJr/Screenshot-2020-10-04-at-6-04-34-PM.png and also wll look like this in light mode https://i.ibb.co/SNJZyCY/Screenshot-2020-10-04-at-6-05-37-PM.png.
So tell me if to retain changes and also checkout if to keep the other buttons style as this or previous one with white floating part.
In this i had made the floating part transparent and according to me it's more attractable than before. If, you say i can keep these changes or i can revert to previous one's.

Neeraj3508: https://i.ibb.co/SNJZyCY/Screenshot-2020-10-04-at-6-05-37-PM.png

Neeraj3508: if i do these changes the white button will like this in light mode https://i.ibb.co/SNJZyCY/Screenshot-2020-10-04-at-6-05-37-PM.png and like this in dark mode https://i.ibb.co/v4WMjJr/Screenshot-2020-10-04-at-6-04-34-PM.png.
Tell me if to keep these changes.
Also check the other buttons, i had changed the floating part to transparent rather than keeping it white. According to me this looks way more cool and attractive. Tell me if to keep these changes or revert back to previous one's.

shahednasser:
I think it's best to have all the button's background transparent as it gives more flexibility to how they can be used. As for how it looks on light and dark mode please don't worry about that.

Neeraj3508:
are u saying the changes as shown in above screenshots as to what you are saying will make all buttons transparent on hover state and there will be no colors, just blank buttons

Neeraj3508:
I think you are talking about the color of part above liquid part. Like in above screenshot. Am i right?

shahednasser:
Yes not the liquid part.

Neeraj3508:
Ok i had done all the changes you asked lemme commit then you can check

Neeraj3508:
@shahednasser  i had done all the changes you asked.

shahednasser:
The "liquid" for white is now transparent but it should be white so that on hover the entire button becomes white (except the text).

Also please remove the word "Button" from the label

Neeraj3508:
Please check @shahednasser

shahednasser:
White does not look good again. There's no animation to it. The button itself should be transparent like the rest but on hover the liquid (which is white) should rise and the button itself will become white.

Neeraj3508:
Actually the transparent part is in repect to buttons background and not the page background so when i am setting transparent for white if can't be seen because that is white transparent tint and if i choose black tint the black button will show same behaviour so either i will have to go any other color tint but that will destroy the beauty of all other buttons.

shahednasser:
Then can we make that change to just white? We need to make the white button work so please find the best way to do it.

Neeraj3508:
Okay @shahednasse

Neeraj3508:
Done @shahednasser. I applied black tint to white button and a border to differerentiate from page

shahednasser:
Great work thanks!

Issue #128, opened 2020-09-15
Add .hover class for force button hover appearance


cosay:
:hover is not trusted attribute so You can't set it programmatically (ex. using JS).
We can add a CSS class to handle this state.
It can be used for button hover animation, for triggering from another hover, etc.

This solution is used, for example, in bootstrap buttons (.active class).

If you think the idea is good, I can work on it. Just assign me  :)

shahednasser:
I'm not sure I understand exactly what the issue is. Can you please elaborate more? Where can the issue be found or what is needed to be changed or fixed

cosay:
This is a **new feature**. I would like to add a css .hover class for each button that has a :hover state.
This will allow you to change the button hover state using Javascript.

cosay:
Example:

To every :hover pseudo-class like this:

```
&:hover {
    animation: pulse 1s infinite;
}
```

I would like to add a class .hover

```
&:hover,
&.hover {
    animation: pulse 1s infinite;
}
```

shahednasser:
I'm not really sure what the purpose of this would be honestly. Plus, this would mean that we would have to add Javascript to the button's functionalities as well (i.e. if someone would use them, they would have to download the css and javascript).

cosay:
No, we just add one class in CSS for each button (.hover class).
I will prepare an example to present what exactly is going on :)

GoodnessOlawoore:
I think I understand this, kindly assign it to me.

GoodnessOlawoore:
> I think I understand this, kindly assign it to me.

Thank you @shahednasser  , I will work on it

shahednasser:
@GoodnessOlawoore any update on this?

cosay:
> @GoodnessOlawoore any update on this?

I'll do it today :)

cosay:
But not yet added for all buttons.
I will try to finish tomorrow :)

Example how to use this feature:
https://jsfiddle.net/cosay/3okb8rs0/

PR #626, opened 2020-10-03
Added .active class to hover and animated buttons

cosay:
New feature #128 

Added .active class for **hover** and **animated** buttons.
Now You can use .active class to force :hover state or controll this state using JS.

Example:
https://jsfiddle.net/cosay/3okb8rs0/

Issue: #128


Zenix27 (approved):

shahednasser:
Hi @cosay I'm still not sure how necessary this feature is. Can you provide some use case for it? I don't think I've seen anyone need it.

PR #631, opened 2020-10-04
Adding Fix for issue of overlapping sidebar #621

Dheerajbana:
Issue:


shahednasser:
This solution doesn't work well because when I decrease the screen size it still overlaps. A better solution in my opinion for now is to set z-index of the footer higher than that of the sidebar when screen width is >= 768px

shahednasser:
Hello @Dheerajbana any update on this?

shahednasser:
Closing this PR due to inactivity on a high priority bug and unassigning issue.

PR #632, opened 2020-10-04
Added new Button "bounce right"

ManideepVakulabharanam:
Added new button in animation type in which button bounces right.


shahednasser:
Thank you for your work @ManideepVakulabharanam but I'm not a big fan of the button. Please next time submit a "new button idea" issue before working on it to get it approved.

Issue #54, opened 2020-09-08
Fix "Fill" rounded buttons on mobile


shahednasser:
On mobile devices (specifically iPhone), when clicking on buttons of type "Fill" the overlayed color doesn't fully fill the container button.

Vansh1067:
can I work on this??

shahednasser:
Sure!

shahednasser:
It seems like no one is currently working on this issue if anyone is interested

reveurguy:
the issue you are stating doesn't seem to exist. Has it been corrected?

shahednasser:
@reveurguy It seems that this is happening to some buttons and not others. Please look at the screenshots for examples
![image](https://user-images.githubusercontent.com/27354907/93749328-6b62e980-fc02-11ea-968c-c3763378a3e9.jpeg)
![image](https://user-images.githubusercontent.com/27354907/93749334-6e5dda00-fc02-11ea-87e9-fdfd9f8cf881.jpeg)

ahmad-shiddiq:
I didn't see that problem honestly, everything works well in my device.

shahednasser:
That's interesting @ahmad-shiddiq what device are you using? Maybe this will help narrow the problem down

ahmad-shiddiq:
I'm sorry, I thought that was IOS. I was trying on Ipad.
![WhatsApp Image 2020-09-25 at 3 07 10 PM](https://user-images.githubusercontent.com/53107961/94337552-6ee8dd00-0015-11eb-951d-6d70ac8444c6.jpeg)

But I tried on my phone too, same didn't see any problem as you said.
![WhatsApp Image 2020-09-26 at 4 30 28 PM](https://user-images.githubusercontent.com/53107961/94337576-b8d1c300-0015-11eb-9c40-48fadb42af3f.jpeg)

shahednasser:
Thank you for sharing that @ahmad-shiddiq . Will keep this issue for future reference then.

samyev:
Hi, has this issue been resolved? if not, can I work on it?

shahednasser:
@samyev it has not been resolved. We've been having trouble reproducing it. If you can reproduce and fix it then go ahead!

Caleb335:
@shahednasser what exactly is the problem?

Are the buttons too round?  On mobile? Please enlighten me, so I can try fixing it

shahednasser:
> @shahednasser what exactly is the problem?
> 
[...]

PR #633, opened 2020-10-04
Fix fill-animation on rounded buttons

basvandenwijngaard:
Fix the white/transparent line on the right of the button for buttons with a fill animation and rounded corners (.fill-color-btn .rounded-btn) on Safari (webkit) on iOs and macOs.

Issue: #54

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -412,7 +412,7 @@ a.yellow-btn:active:not(.fill-color-btn) {
   border-radius: 25px;
 }
 .aurapulse-btn:hover:before {
-  animation: aurapulse 0.4s infinite;
+  animation: aurapulse 0.5s infinite;
 }
 @keyframes aurapulse {
   0% {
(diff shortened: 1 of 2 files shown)


shahednasser:
That did fix the issue wow! but for some reason the animation looks slow now, can we make it faster or was that the reason why it was breaking?

basvandenwijngaard:
> That did fix the issue wow! but for some reason the animation looks slow now, can we make it faster or was that the reason why it was breaking?

It's more like a small work-around for the issue (that seems to only occur in Safari). Just before the animation on the `:before` is about the end I change the button background color to the same color, only when it's a rounded button.

I didn't change the original animation, but maybe because the button now had a second animation on de background, it slows down a bit. I didn't notice it at first, but the animation does seems to be a little bit slower than before.

I'm not sure, but to me is seems to be more the same speed as the non-rounded buttons.

shahednasser:
@basvandenwijngaard the weird thing is when testing it on the laptop it looks fine but then on my phone it looks a bit laggy. But I think I'll take this fix for now. Thank you so much!

shahednasser:
@all-contributors please add @basvandenwijngaard for code and bug

PR #677, opened 2020-10-05
Glossy button.

adsingh14:
Added Glossy button under Special Category.

Issue:


src/components/special/_glossy-btn.less
```diff
@@ -0,0 +1,30 @@
...
+    top: 0;
+    height: 25%;
+    background: rgba(white, 0.4);
+    left: 0;
+  }
+  &.white-btn:after {
+    background: rgba(#ddd, 0.4);
+  }
+
+  &:active:after {
```
shahednasser, line 26: nest this inside `&:active` on line 9 making the necessary adjustments to the selector

shahednasser (changes requested):
I like it but maybe the top part of the button should be reduced? Also, add some transition to the button for the mouse in/out effects to look more seemless. And try to make the box-shadow color lighter.

adsingh14:
> 
> 
> I like it but maybe the top part of the button should be reduced? 
I was thinking to increase the top part. Anyways, I will reduce it.

adsingh14:
#682 Updated.

PR #679, opened 2020-10-05
Refactored appstore

SamD97:
Issue: Common in properties in `&.appstore` and `&.playstore`

Solution: Common properties were moved out to the parent.


shahednasser:
Regarding the comment you mentioned I didn't notice the different left attributes, so put those with the content instead of in the parent element.

SamD97:
> Regarding the comment you mentioned I didn't notice the different left attributes, so put those with the content instead of in the parent element.

I am slightly confused, you meant to say "I did notice the different `left` attributes" right?
https://github.com/sButtons/sbuttons/blob/82dfc98c10f6020f1797b537b5cba436abd7333b/src/components/icon/_appstore.less#L46
https://github.com/sButtons/sbuttons/blob/82dfc98c10f6020f1797b537b5cba436abd7333b/src/components/icon/_appstore.less#L69

Anyways I am assuming that's what you meant based on the latter part of your reply and undo that change.

shahednasser (changes requested):
So I checked the current website and apparently the different left attributes (100px and 87px) is only applied when the button is block, else it's always 50px so can we refactor the code to portray that? Change the 87px to 102px as a fix to the original button's design.

Also for the `.appstore-btn` when it's a block button please change the padding left to 30px as a fix to the original button.

dist/sbuttons.css
```diff
@@ -1924,10 +1924,12 @@ a.yellow-btn:active:not(.fill-color-btn) {
 }
 .appstore-btn {
   box-shadow: 0px 1px 4px #4d4d4d;
+  background-color: #000000;
+  color: #fff;
   transition: all 0.2s;
   position: relative;
   padding-top: 15px;
-  padding-left: 60px;
+  padding-left: 30px;
```
SamD97, line 1932: The `padding-left: 30px` only affects `.block-btn` because of line 31

src/components/icon/_appstore.less
```diff
@@ -1,9 +1,11 @@
 .appstore-btn {
   box-shadow: 0px 1px 4px #4d4d4d;
+  background-color: #000000;
+  color: #fff;
   transition: all 0.2s;
   position: relative;
   padding-top: 15px;
-  padding-left: 60px;
+  padding-left: 30px;
```
shahednasser, line 8: This change should only be applied to `applestore.block-btn` not all. Please fix this and change the default back to 60px

shahednasser (changes requested):

src/components/icon/_appstore.less
SamD97, line 8: The `padding-left: 30px` only affects `.block-btn` because of line 31 hence for other buttons it is `50px`

shahednasser:
Great work thank you!

SamD97:
Thank you for all the help and feedback :+1:

PR #683, opened 2020-10-05
add jump layer button

printf-ana:
Issue: I add a new type of button where, when the user pass the mouse in, the button jump in layers.


shahednasser:
Thank you @printf-ana for the contribution but unfortunately, I don't this idea fits in our project.

PR #711, opened 2020-10-07
Added Flip button

vsaini212223:
New button added


shahednasser (changes requested):
Hi @vsaini212223 thank you for your work. I think the button would look better if the flip was more 3D. You can check this as an example: https://codepen.io/robbue/pen/hJDlA

vsaini212223:
Yeah, 3D looks more awesome.
Ok... I will make it 3D as soon as possible
Thanks for the suggestion.

shahednasser:
Hi @vsaini212223 any update on this?

Issue #564, opened 2020-10-02
Sidebar overlays footer


arjuna1729:
**1 - Describe the bug**
The sidebar list overlays footer when scrolling to the bottom of the page.
**2 - Expected behavior of what should have happened (if applicable)**

**3 - Screenshots (if applicable)**
![image](https://user-images.githubusercontent.com/33598813/94926378-50d02080-04de-11eb-9997-530dbc0519e7.png)

**4 - Additional notes**

pinkman7009:
can i work on this issue?

Dheerajbana:
can i take thiis up

shahednasser:
Hi @arjuna1729 can you provide more info on this? What browser are you using and what's the screen width that this bug is happening on? Thanks!

adsingh14:
I am using Mozilla Firefox. I think that's happening in all browsers because it's the CSS syntactical error. 

![image](https://user-images.githubusercontent.com/35252877/94959985-bc7eb180-050f-11eb-8436-2ff050729ad6.png)

shahednasser:
You are all correct this is due to an increase in the number of items, not a CSS syntactical error. It can be resolved in multiple ways.

@arjuna1729 thank you for reporting the bug. Would you like to work on it or should I assign it to someone else?

pinkman7009:
can i work on this?

shahednasser:
you may go ahead @pinkman7009

adsingh14:
Footer has to be changed. Need to fill it up with more content.

pinkman7009:
thank you @shahednasser . i will start working on this,

shahednasser:
@adsingh14 please start a new issue with any idea you have instead of commenting on unrelated issues.

pinkman7009:
@shahednasser i am not getting this error on chrome and firefox at 100% width screen size. can you tell me where i might be going wrong?

shahednasser:
This could be due to different screen sizes. What's your screen width?

pinkman7009:
@shahednasser my screen size is 15.6 inch which is around 1450 pixels.

shahednasser:
That's odd but everyone has seen the bug by now. If you can't reproduce it then maybe we should assign it to someone else. Please let me know.

pinkman7009:
@shahednasser okay then i guess you should assign it to someone else.

shahednasser:
@Dheerajbana do you still want to work on this?

Dheerajbana:
[...]

PR #719, opened 2020-10-07
fix overlaying scrollbar

Caleb335:
Issue: #564 

@shahednasser 

At first, when I launched the website on my PC, The links were not overlapping the footer, that was because the content wasn't much then, so I went on to add additional button examples in the `button-examples.js` file so I'd be able to see the error.

I fixed this error by adding a fixed width to the submenu links of all the button examples and allowing them to overflow, thus, the scrollbar you see..

Kindly check the GIF below to see it in action.

ezgif com-gif-maker (3)

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -309,6 +309,8 @@ footer {
   padding-top: 20px;
   padding-bottom: 20px;
   background-color: #e6e6e6;
+  z-index: 2;
+  position: relative;
 }
 
 .tagline {
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -2,7 +2,7 @@
   display: flex;
   flex-direction: column;
   background-color: white;
-  z-index: 10;
+  z-index: 1;
 }
 .sidebar.scrolling {
   position: fixed;
@@ -36,6 +36,8 @@
 }
 .sidebar .submenu-links {
   margin-left: 10px;
+  height: 135px;
+  overflow-y: auto;
 }
 .sidebar .submenu-links:not(.show) {
   display: none;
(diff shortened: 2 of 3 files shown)


shahednasser (changes requested):
That's a nice idea but on certain screen widths the buttons still overlap with the footer (for example when the height is less than 600px). So can you still make the sidebar hide behind the footer? This is just a temporary change so I don't want you to work on a complex fix.

Also, can you make all these changes (the changes you made and making the sidebar behind the footer) only on devices with width >= 768px? As on small devices it is unnecessary

Caleb335:
> That's a nice idea but on certain screen widths the buttons still overlap with the footer (for example when the height is less than 600px). So can you still make the sidebar hide behind the footer? This is just a temporary change so I don't want you to work on a complex fix.

Okay, I'd work on that

> Also, can you make all these changes (the changes you made and making the sidebar behind the footer) only on devices with width >= 768px? As on small devices it is unnecessary

Oh! I noticed that also when I was viewing the preview

Caleb335:
@shahednasser I've added the requested change.

Kindly review

shahednasser (changes requested):
Looks great but these changes are still being applied to small devices. Can you add the media query to avoid that? All changes made should be in the media query.

Caleb335:
> Looks great but these changes are still being applied to small devices. Can you add the media query to avoid that? All changes made should be in the media query.

All done now.

Kindly review, Thanks

shahednasser (changes requested):
I just realized that the links are now not clickable, so can you change the z-index of sidebar to 0 and add the following CSS rules to footer:

```
z-index: 2;
position: relative;
```

Caleb335:
@shahednasser, I've made changes to the files you requested.

The links are clickable now.

Please review

shahednasser:
Great thank you!

@all-contributors please add @Caleb335 for bug

Issue #655, opened 2020-10-05
Refactor code in `_glow.less`


shahednasser:
In `src/components/animated/_glow.less` there are two mixins `glowBorderRule` and `glow`, then inside each button color (for example `blue-btn`), the following code is used:

```
    transition: all 0.5s ease;
    .glowBorderRule(@neonBlue);
    &:hover {
      .glow(@neonBlue);
    }
```

with `@neonBlue` replaced with the color needed for the button.

We need to create a new mixin with the above code that takes the color as a parameter, then use it accordingly throughout the file.

Also, place all mixin declarations at the top of the file.

Hisime:
Hello. Please assign it to me.

shahednasser:
Sure @Hisime go ahead

PR #722, opened 2020-10-07
Refactor glow less

Hisime:
Issue: https://github.com/sButtons/sbuttons/issues/655


src/components/styles/_glow.less
```diff
@@ -1,88 +1,51 @@
-.glowBorderRule(@color) {
-  border-color: @color;
-  background-color: rgba(@color, 0.2);
-}
-
 .glow-btn {
-  transition: all 0.5s ease;
   border: 2px solid;
   color: @black;
```
shahednasser, line 3: Can you change this to `@darkText !important;` ? I know it was like this in the original code but it would be a good fix to it.

shahednasser (changes requested):

src/components/styles/_glow.less
Hisime: Done

shahednasser:
Great thank you!

@all-contributors please add @Hisime for code

PR #725, opened 2020-10-08
Grammatically updated the subtitle in index.html

drewren25:
I made a grammatical change as well as a change in word choice for the subtitle in index.html. Originally it said "Simple buttons you can use easily for your next project." and I changed it to "Easy to use buttons you can use for your next project!"
I found it a bit wonky to read with both "simple" and "easily" in the same sentence so I changed it to just say "easy to use" which I found more concise without losing the original meaning.
I also added an exclamation mark to add a bit of enthusiasm on our part as well as try to instill some into possible users. I was choosing between an exclamation mark and just using no punctuation at all, but I just thought a period was too stern and emotionless for something that I think deserves excitement.

### DIFF
--- a/index.html
+++ b/index.html
@@ -44,7 +44,7 @@
     <!-- seo friendly og tags-->
     <meta
       name="description"
-      content="Easy to use buttons you can use for your next project!"
+      content="Simple to use buttons you can use for your next project!"
     />
     <meta
       name="keywords"
@@ -53,7 +53,7 @@
     <meta name="og:url" content="https://sbuttons.github.io/sbuttons/" />
     <meta
       name="og:description"
-      content="Easy to use buttons you can use for your next project!"
+      content="Simple to use buttons you can use for your next project!"
     />
     <meta name="og:image" content="site logo url" />
     <meta name="og:type" content="website" />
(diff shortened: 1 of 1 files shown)


shahednasser:
I see what you mean. How about we use Simple instead of Easy?

drewren25:
> I see what you mean. How about we use Simple instead of Easy?

Sounds good to me!

drewren25:
Sorry this is my first time working on an open source project like this, does that mean I should make that edit?

shahednasser:
No problem! Yes you should make the changes in your branch then commit the changes, it will update this PR automatically.

drewren25:
> No problem! Yes you should make the changes in your branch then commit the changes, it will update this PR automatically.

Oh ok I see, I'll do that by the end of today(pst)!

drewren25:
Just sent in the commit, is this what you were looking for?

shahednasser:
After looking into this more I think we should keep the current subtitle. I appreciate your contribution and I'll label this as "hacktoberfest-accepted". Thank you.

drewren25:
No problem, thanks for helping me make my first contribution! One more question, since this pr is closed github asks if I want to delete this branch and/or delete this fork. Should I?

shahednasser:
@drewren25 sure you don't need it anymore.

Issue #750, opened 2020-10-09
Add active state for star button


shahednasser:
Currently the star icon button on `:active` gets colored yellow, however it is more convenient to also add an active state for the button (a new class `.active`) where the color of the star will be yellow. It should also have a nice transition between the inactive and active state. It shouldn't be displayed on the website, but if someone adds the class `.active` to the button or removes it, the transition should look nice.

crew-guy:
Can I do it please ? I just have 1 small other pending issue

shahednasser:
@crew-guy I really appreciate you contributing to this repository but I can't give you too many issues to work on at the same time. Once you finish at least one of them you can then get assigned to a new one. Thank you!

pontakornth:
@shahednasser I already forked the repository but I forget to add comment here. Could you assign this to me?

shahednasser:
sure @pontakornth but please follow the other instructions I gave you as well.

crew-guy:
> @crew-guy I really appreciate you contributing to this repository but I can't give you too many issues to work on at the same time. Once you finish at least one of them you can then get assigned to a new one. Thank you!

Yaya cool ! Guess I am taking wayy too much work to properly process and finish it all 😂✌🏻

PriyanshTri:
can you assign this to me? thanks

pontakornth:
> sure @pontakornth but please follow the other instructions I gave you as well.

Done. The check is still failed and I have no idea why.

maityankan10:
i would like to contribute

PR #754, opened 2020-10-09
Add active state for star button (#750)

pontakornth:
I add the active state for star button with mixin so both `:active` and `active` class can have same style.
Issue: 750


shahednasser:
Hi @pontakornth please first comment on the issue to be assigned to it. Also, you need to make sure your repo is up to date and create a new branch instead of using the master branch. Thank you.

PR #764, opened 2020-10-09
Stack buttons issue #729 resolved as per PR #746

crew-guy:
Stack Buttons now :

1. Are added under Button Styles 
2. Are using colors from a src/components/_variables.less file 

as discussed and reviewed in #746

Issue #729, opened 2020-10-08
[BUTTON IDEA]: 3D Layered Icon Button


crew-guy:
**1. Name of button**: 3D Layered Icon Button

**2. Description**: Initially skewed button. On hover, we see a decreasing transparency 3D stack below button

**3. Button Type (Animated, Special, etc...)**: Hover button

**4. Will you work on it?**: Yes. I'll try adding some fill-in or ripple effect to complement the effect even better. I also want to implement different style for dark and light theme for buttons of different platforms like Google, Facebook, Instagram, Snapchat, LinkedIn, Github and more.

---

![maxresdefault](https://user-images.githubusercontent.com/69717009/95454186-bb270c00-0989-11eb-9c8c-69041b1dcad2.jpg)

shahednasser:
@crew-guy someone tried implementing this before, but it didn't look so good. You may work on it and I'll see how it looks in the end.

crew-guy:
> @crew-guy someone tried implementing this before, but it didn't look so good. You may work on it and I'll see how it looks in the end.

You were so SO right about this Ma'am. It took me forever to implement just a rudimentary version of this (check out pull request #746 ) 😅😁

But it was really fun ! I look forward to coming up with and implementing more such design ideas !

shahednasser:
You did a great job at it @crew-guy , I just need you to make the changes mentioned in the PR and I'm sure it will be a part of the repo soon! Thank you for contributing and trying hard.

crew-guy:
Hey !! I have completed both tasks assigned to me. Check out PR - #764 and #763  
I gotta say, refactoring my own code from icon buttons to button styles and that too whilst minimizing CSS lines has gotta be the one of the coolest CSS problems I've had to crack
I honestly had no clue about PRs and branches but thanks to these assignments, I have a fairly good idea of how to work around them.
You're a great mentor Ma'am and I really enjoy working with you ! 🙏🏻😄

shahednasser:
@crew-guy thank you so much and I hope this repository keeps helping you learn new things! I'll check your PRs soon.

crew-guy:
Thanks !!🙏🏻

PR #769, opened 2020-10-10
Added stack buttons #729 as per #764

crew-guy:
1. Added some website layout changes (to fit this button container without overlap or other layout issues) in assets/css/index.css instead. This applies to all rules related to #stack
2. Placed this in the correct alphabetical order under "Button Styles" and changed label to just "Stack"
3. Removed "./" in several imports as it is unnecessary
4. Really tried cracking my head around trying to basically make the text inside the button element instead. I know putting it in :before makes it hard to customize (as the user will need to use whatever text they have). I've been at this since hours ( diverting text from index.js to fit inside before element of button ) and I am no closer to actually implementing it. Can you please help me resolve this ? There must be some way to work around this exception right ?


shahednasser (changes requested):
Looks great but please remove changes in README.

crew-guy:
> Looks great but please remove changes in README.

Ma'am I swear I never go into README or any other file when working on the buttons. Yet somehow, I always get changes in them. Is there a way to control it ?

shahednasser:
Can you just copy the content of the current readme, paste it in the file in your repo and commit it? I'm not sure what is causing it but maybe some ide you're using.

Issue #416, opened 2020-10-01
Colored Background on Hover Button


shahednasser:
**1. Name of button**: Colored Background on Hover Button

**2. Description**: A button that is originally transparent but its background becomes colored on hover

**3. Button Type (Animated, Special, etc...)**: Button Styles

**4. Will you work on it?**:

sindhwanicheshtha06:
I can work on this

yaspalsingh115:
Yes, I can do.

LuciKritZ:
I can work on it too.

shahednasser:
assigning @sindhwanicheshtha06  for replying first

crew-guy:
Hey ! Does anyone need any help with this ?

shahednasser:
Unassigning this due to inactivity so that someone else can work on it.

shahednasser:
@crew-guy if you finish the stack buttons and no one is assigned I'll let you know.

developerzone:
I can work on this.

shahednasser:
Sure @developerzone go ahead

developerzone:
I am getting error while creating a PR. Please check this at https://github.com/sButtons/sbuttons/pull/776 and kindly guide me on this if I am doing anything wrong.

maityankan10:
can I work on this

HERITAGEALABI:
> **1. Name of button**: Colored Background on Hover Button
> 
> **2. Description**: A button that is originally transparent but its background becomes colored on hover
> 
> **3. Button Type (Animated, Special, etc...)**: Button Styles
> 
> **4. Will you work on it?**:

yes i can work on it

shahednasser:
Hi @HERITAGEALABI please do. I'll assign you.

HERITAGEALABI:
@shahednasser i apologize i will work on a button type

venkataraman123:
I can work on it

mrSidSat:
Can I work on this issue

shahednasser:
@mrSidSat you may work on this. I'm changing the priority of this button to High as it has taken too long and I think it's an important button

venkataraman123:
Can I work on this

rifafatima:
Yes, I can

shahednasser:
@rifafatima since this is a high priority issue you can work on it as well. We need to resolve this one soon.

rifafatima:
> @rifafatima since this is a high priority issue you can work on it as well. We need to resolve this one soon.

[...]

PR #777, opened 2020-10-10
Develop special button

developerzone:
1. Worked on issue #416 
2. Added hover on transparent handwritten buttons under special buttons


shahednasser:
This PR does not relate to the issue. Please review the issue again. We just want a button with a transparent background (like outline buttons) that on hover has the background color of the outline.

PR #785, opened 2020-10-10
Changed glossy button border

drewren25:
I added a "border: none" to the glossy-btn class because I noticed the glossy buttons had a border that cut off the glossy white part of the button. When I saw them for the first time that part kind of threw me off so that's why I changed it to this.


shahednasser:
Thank you for noticing that!

@all-contributors please add @drewren25 for code

Issue #765, opened 2020-10-09
[BUTTON IDEA]: Hamburger Menu Button


senguyen1011:
**1. Name of button**: Hamburger Menu Button

**2. Description**: Hamburger menu buttons/icons are typically three horizontal lines that toggle the appearance of an off-screen menu. The humburger menu button when toggled will change to a different icon, such as an 'X' to indicate clicking it again will close the menu and reset the button.

**3. Button Type (Animated, Special, etc...)**: Icon Buttons

**4. Will you work on it?**: Yes

![image](https://user-images.githubusercontent.com/52718953/95587188-bb69ea80-0a0f-11eb-8b4b-c6dda77a9ac4.png)

senguyen1011:
I'd like to work on this for Hacktoberfest, and possibly add animated variants of the hamburger icon. Please let me know if I didn't create this issue correctly.

shahednasser:
We're currently not adding icon buttons but if you're gonna add it animated as you say (change between hamburger and X) then you may. But here are a few notes:

All buttons are implemented without JS, only CSS. So you need to do things:

1. Create the hamburger button which is the original state.
2. Add an `.active` state on that button that makes the hamburger button an X button. The transition between them needs to be clean.
3. Display both states on the website but without the actual toggling as it will require JS and we're not doing that for now.

Let me know if you can work on it. I can help you with anything you need as well.

senguyen1011:
Yes, I can create the button and the active state. If you create an issue, please assign it to me, thank you.

PR #787, opened 2020-10-11
Add hamburger button

senguyen1011:
Added a hamburger icon button. Animations are included in CSS, but require JS to toggle between hamburger (default) state and active state. Active states include a cross (X), minus sign, left arrow, and right arrow.

Issue: #765


src/components/icon/_hamburger.less
```diff
@@ -0,0 +1,68 @@
...
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    background: currentColor;
+    transform: translate(-50%, -50%);
+    transition: transform 0.2s ease;
+  }
+}
+
+.hamburger-btn:hover {
```
shahednasser, line 28: nest inside `.hamburger-btn` in the beginning.

shahednasser (changes requested):
Thank you for your work! I'm not sure why you have many is-active states. I think you just need the x one and the transition should just be by adding `.is-active` (without adding `hamburger-x-btn`). Also the transition between hamburger and x doesn't look so clean so please work on it.

src/components/icon/_hamburger.less
senguyen1011: `.hamburger-btn` is a parent class and other classes like `.hamburger-x-btn`, `.hamburger-min-btn`, `.hamburger-larr-btn` determine what icon the button will transform into when toggled. Also, could you tell me what issue you're finding with the transition? When I test it locally, it seems fine to me (there are 2 lines, :before and :after, hidden in the middle bar of the hamburger icon. They both turn 45 degrees to form the 'X' while the top and bottom bars fade away)
hamburger

shahednasser: I understand but I think the most used is hamburger to x so let's keep it simple by just adding those two states. By just adding `.active` to `.hamburger-btn` it should change to the x. As for the transition I think it looks good so forget my comment about it.

senguyen1011:
I've removed the variant hamburger icons, but my checks aren't passing and I'm unsure of how to fix that.

shahednasser:
Great work thank you!

@all-contributors please add @senguyen1011 for code and idea

Issue #794, opened 2020-10-12
Sidebar overlapping header


shahednasser:
After fixing sidebar overlapping footer, now sidebar overlaps header. I suggest going through `assets/js/index.js` as the solution to this is most probably there.

wbrijesh:
Can i do this using z-index?
Having the headers z-index higher than that of sidebar might solve your problem

krishnasreekoganti:
Hi can I work on this issue?

PR #795, opened 2020-10-12
Fix/sidebar overlapping

michalwowk:
I've removed unnecessary JavaScript and CSS code and fixed the problem with `position: sticky;`

Issue: Sidebar overlapping header #794.

### DIFF
--- a/assets/css/sidebar.css
+++ b/assets/css/sidebar.css
@@ -3,7 +3,7 @@
   flex-direction: column;
   background-color: white;
   position: sticky;
-  top: 3rem;
+  top: 90px;
 }
 
 .sidebar a {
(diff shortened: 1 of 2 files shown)


michalwowk:
@shahednasser have you tested that? I just remove unnecessary JavaScript code, nothing else. We don't need extra "scrolling" class or scroll detection if we are using `position: sticky`. I've just tested that (Safari, Chrome, Firefox) and it works.

michalwowk:
@shahednasser  Oh sorry, I didn't notice this navbar. Fixed.

shahednasser:
Great thank you!

@all-contributors please add @michalwowk for code and bug

PR #799, opened 2020-10-13
Added _whatsapp.less  in social

harsh1868:


shahednasser (changes requested):
Hi @harsh1868 thank you for contributing. I have some notes about your work:

1. You're giving the button a background color of `#000000` but the background color should be whatsapp's green color
2. You need to import `_whatsapp.less` in `src/components/_social.less`
3. You need to add the button in `assets/js/buttons-examples.js` under Social
4. (optional) you need to run `npm run compile` to see the changes.

Please make the changes listed above.

harsh1868:
@shahednasser   i have changed as per your suggestion please review it.

shahednasser:
There seems to be a problem with the compilation so can you please run `npm run compile` and commit the changes?

shahednasser (changes requested):
You still need to run `npm run compile` and commit the changes

harsh1868:
@shahednasser I tried npm run compile and committed changes. please review .

shahednasser (changes requested):
All looks good but the whatsapp block button has the text and the icon overlapping. Please fix that.

harsh1868:
> All looks good but the whatsapp block button has the text and the icon overlapping. Please fix that.

Yes, but its like more inbuilt issueof that block button. Or please suggest me a way at least ...

shahednasser:
I understand @harsh1868 but you can make changes specific to the whatsapp button by adding CSS rules to `.whatsapp.block-btn`. I did a little fix and for example you can set margin-right for `.whatsapp.block-btn:after` to 0. You might need to use `!important` so please try it out.

harsh1868:
> I understand @harsh1868 but you can make changes specific to the whatsapp button by adding CSS rules to `.whatsapp.block-btn`. I did a little fix and for example you can set margin-right for `.whatsapp.block-btn:after` to 0. You might need to use `!important` so please try it out.

Tried this but not working...

shahednasser:
Make sure to run `npm run compile` after every change before testing it. The changes will not appear on the website without compiling.

harsh1868:
@shahednasser  please review

src/components/social/_whatsapp.less
```diff
@@ -0,0 +1,12 @@
+.social-btn {
+  &.whatsapp {
+    background-color: #25d366;
+    color: #fff;
+
+    &:after {
+      color: white;
+      right: 0px !important;
```
shahednasser, line 8: this should be inside `.whatsapp.block-btn:after` something like this:
```
&.block-btn {
&:after {
right: 0px !important;
}
}
```

harsh1868:
@shahednasser  finally i think now its done...

src/components/social/_whatsapp.less
harsh1868: Please check and review..

shahednasser:
Great thank you!

@all-contributors please add @harsh1868 for code

PR #845, opened 2020-10-17
new animated-button (Hover)

Erxical:
adding new animated button style name Hover

Issue:


shahednasser:
Thank you for your work but this is similar to the key button and I find it unnecessary to have duplicates. I'll label this "hacktoberfest-accepted" since I appreciate your work.

PR #860, opened 2020-10-18
Trash Button added

sayalitandel:
Issue:


sayalitandel:
Hey @shahednasser! This is the second PR of my life, I have added a trash button in the Icon Button category. I would be glad if you would review it. Thank you.

PR #870, opened 2020-10-20
key button updated

I3han:
Key Button Updated

Issue: Improve Key Button

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -463,3 +463,7 @@ button.clipboard:focus {
   padding-bottom: 4rem;
   margin-bottom: 4rem;
 }
+
+.key-btn {
+  font-family: "Syne", sans-serif;
+}
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -465,5 +465,5 @@ button.clipboard:focus {
 }
 
 .key-btn {
-  font-family: "Syne", sans-serif;
+  font-family: "Syne", sans-serif !important; 
 }
--- a/src/components/styles/_key.less
+++ b/src/components/styles/_key.less
@@ -2,17 +2,17 @@
   border-top: 2px solid @color;
   background-color: @color;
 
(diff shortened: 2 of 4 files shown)


src/components/styles/_key.less
```diff
@@ -1,11 +1,18 @@
 .btnVariant(@color, @keyHoverColor, @keyBoxColor) {
   border-top: 2px solid @color;
   background-color: @color;
-  box-shadow: inset 0px -12px @keyBoxColor;
+
+  -webkit-box-shadow: inset 0 1px 0 #ffe5c4, 0 10px 0 @keyBoxColor;
```
shahednasser, line 5: instead of `#ffe5c4` please use `@color`. Replace it everywhere in this mixin.

src/components/styles/_key.less
```diff
@@ -21,16 +28,24 @@
...
   border-radius: 4px;
   position: relative;
-  border-bottom: 6px solid @cola;
-  border-left: 2px solid @cola;
-  border-right: 2px solid @cola;
 
   &:active {
     outline: none;
     top: 2px;
     border-top: 2px solid @cola;
```
shahednasser, line 35: Remove this

shahednasser (changes requested):
Looks great I just have a few comments below.

Also, I'm not a fan of the font. Can you add a rule in `assets/css/index.css` for `.key-btn` to use "Syne" font instead?

assets/css/index.css
```diff
@@ -463,3 +463,7 @@ button.clipboard:focus {
   padding-bottom: 4rem;
   margin-bottom: 4rem;
 }
+
+.key-btn {
+  font-family: "Syne", sans-serif;
```
shahednasser, line 468: It seems that this needs `!important` to work properly so please add it.

shahednasser:
Great work thank you!

@all-contributors please add @I3han for code

PR #891, opened 2020-10-22
Website Revamp

HADES-01:
## Website Revamp
This PR marks the second phase of the Website Revamp which consists of the following changes :
1. Structural CSS changes i.e. different files for different components like header, main, footer etc. for easy understanding.
2. Style changes for sidebar and footer.
3. More vibrant color scheming.


shahednasser (changes requested):
@HADES-01 thank you for the great work! I have a few remarks below:

1. I think the website's color scheme and style is now all coordinated, except for the background color of the footer. Let's change it to something that matches the rest of the website. Maybe light gray? You pick what you find most fitting.
2. When the screen's width is less than 992 and is greater than 768 the sidebar overlaps with the content of the website. This is due to the fact that our code makes the sidebar hidden when the screen is less than 768. Due to the website changes, I think we need to change it now to make the website hidden at 992 instead.
3. There seems to be an issue with the sidebar and footer in smaller screens. When the screen is less than 992 and greater than 768 the sidebar is hidden behind the footer and the footer is not at the end of the screen, leaving a white space in the end
sButtons-Simple-and-Creative-Buttons (1)
4. On devices less than 768 when the sidebar is hidden, there's also some white space that's covering the footer.
sButtons-Simple-and-Creative-Buttons (2)

There are also some content issues that I found, like the link to the examples section in the footer takes to the repo rather than the webpage, but that's ok let's leave them for now. We can create issues later for beginners to fix them.

Thank you for your great work I appreciate it!

HADES-01:
### Requested Changes
- [x]  I think the website's color scheme and style is now all coordinated, except for the background color of the footer. Let's change it to something that matches the rest of the website. **Made it a similar shade of the main color**  
- [x]  When the screen's width is less than 992 and is greater than 768 the sidebar overlaps with the content of the website. This is due to the fact that our code makes the sidebar hidden when the screen is less than 768. Due to the website changes, I think we need to change it now to make the website hidden at 992 instead.
- [x]   There seems to be an issue with the sidebar and footer in smaller screens. When the screen is less than 992 and greater than 768 the sidebar is hidden behind the footer and the footer is not at the end of the screen, leaving a white space in the end.  
- [x] On devices less than 768 when the sidebar is hidden, there's also some white space that's covering the footer.

### Other Changes
-[x] Put all colors in `variables.css` for more accessibility.

shahednasser:
It all looks great there's just one final change. Please add to `.sidebar-list, .instructions-sidebar` the following:

```
overflow-x: hidden;
```

The reason behind that is that the horizontal scrollbar was showing for some devices. 

And remove the following for `.sidebar a`:

```
white-space: nowrap;
```

That way it won't exceed the width of the sidebar

shahednasser:
Ok I think everything looks great! Thank you so much for your work I appreciate it. I'll merge this to the master branch as well.

PR #898, opened 2020-10-22
Button - View on on Github

HERITAGEALABI:
Added transparent background and colored hover to the "Special " Button type

Issue: #416


shahednasser:
Hello @HERITAGEALABI I think you misunderstood what was asked in the issue. We need to create a new button type that on hover, its background is colored. Let me know if you can do that or not.

PR #900, opened 2020-10-22
Do not launch the button examples script on the "Get Started" page

glebkema:
I noticed that `button-examples.js` is also called for `getStarted.html`. This script makes up the sample buttons for the home page, and as far as I'm concerned, it's not needed on the "[Getting Started](https://sbuttons.github.io/sbuttons/getStarted.html)" page.

At the same time a simple measurement shows that the script runs for about 500 ms for each page load:
```js
$(document).ready(function () {
    start = Date.now();
    // original code is here
    console.log( 'Unclaimed work lasted', Date.now() - start, 'ms' );
});
```
Due to this I've made two changes:

1. Commented out the call to the script. Now it is obvious that the script is not involved in this file.
2. Transfered to the beginning of the script the initialization of the pointers to the sidebar and to the block of the buttons. If one of them was not on the page, then I interrupt the script. The script will now stop even if it is run by mistake.

After disabling the script, the chrome developer tools show quite a significant improvement in page load performance.

block-unused-scriot-perfomance

On the main page, the script works the same as before.

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -59,7 +59,7 @@ $(document).ready(function () {
           label: "Gradient",
           classes: "gradient-btn",
         },
-       
+
         {
           type: "xsmall",
           label: "Extra-small",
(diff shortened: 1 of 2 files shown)


shahednasser:
Thank you so much for the detailed work I appreciate it! Hope you help out with our repository more often.

@all-contributors please add @glebkema for code

PR #919, opened 2020-10-23
Added a button that curls on hover

wisetok12680:
Added a button that curls on hover

Issue:
No Issues

### DIFF
--- a/src/components/animated/_papercurl.less
+++ b/src/components/animated/_papercurl.less
@@ -1,0 +1,31 @@
+.papercurl-btn {
+  backface-visibility: hidden;
+  box-shadow: 0 0 1px rgba(0, 0, 0, 0);
+  display: inline-block;
+  position: relative;
+  transform: translateZ(0px);
+  vertical-align: middle;
+
+  &:before {
+    background: rgba(0, 0, 0, 0)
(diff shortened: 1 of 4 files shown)


src/components/animated/_papercurl.less
```diff
@@ -0,0 +1,31 @@
...
+    position: absolute;
+    right: 0;
+    transition-duration: 0.3s;
+    transition-property: width, height;
+    width: 0;
+    border-color: white;
+  }
+  &:hover::before,
+  &:focus::before,
+  &:active::before {
```
shahednasser, line 27: You can nest before inside as well. The code should be something like this:

```
&:hover,
  &:focus,
  &:active {
   &:before {
      //code here
    }
}
```

shahednasser (changes requested):
You've done the nesting well! I just need you to make this last change

src/components/animated/_papercurl.less
wisetok12680: Ok

wisetok12680: I just made the requested change.
Hope everything is alright now.

shahednasser:
Great work thank you!

@all-contributors please add @wisetok12680 for code and idea

PR #940, opened 2020-10-24
Fix #862. Show all levels in the sidebar menu

### LINKED ISSUE #862: While scrolling the Base Icon Button Section after Base Button the other two button and their sub section are not visible .
**Bug Description**
    While scrolling on getStarted.html page in Base Icon Button Section after Base Button the other two button are not visible or 
    becomes hidden on further scroll after base button and also its sub section are also not visible while scrolling down.

Screenshot 2020-10-19 002231

**For Base Button Working fine but its sub section are not visible while scroll.**

Screenshot 2020-10-19 002315

**But for Rounded Button and Block Button there is no base icon button section  visible in side navbar just after the base button.**

glebkema:
Fix #862. 

Commits:
1. Show all accessors instead of the parent only.
The key point is to make visible the entire chain of nested blocks in a menu of 3 or more levels.

2. Rename auxiliary variables so that their meaning becomes clear even without analyzing the HTML layout.

3. Reorganize the "if else" block to make its logic more obvious.

4. Remove the unused `return false;` statement:
    - the function has no other `return` instructions;
    - without a `return` statement, the function returns the value` null`, which plays almost the same role in comparison operations as `false`;
    - the function is called from two places in the code without any interest in its result.

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -46,7 +46,6 @@ $(document).ready(function () {
           } else {
             $(".submenu-links.show").removeClass("show");
           }
-          return false;
         }
       }
     });
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -34,9 +34,9 @@ $(document).ready(function () {
         var sidebarLink = $(".sidebar a[href='#" + $(this).attr("id") + "']");
         $(".sidebar a.active").removeClass("active");
         sidebarLink.addClass("active");
-        var hasClass = sidebarLink.hasClass("submenu-link"),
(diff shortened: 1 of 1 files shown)


shahednasser:
Great work thank you!

@all-contributors please add @glebkema for code and bug

shahednasser:
@glebkema thank you for always providing good work. I would like to add you as one of this Repo's reviewers. You will be able to review/close issues as well as PRs. I'm trying to add more members due to my busy schedule. If you would like to be a part of our organization I hope you will accept the invite!

PR #941, opened 2020-10-24
Fix the duplicate `block` id on the `Get Started` page

glebkema:
Currently the subtitle in "Usage" and the subsection in "Base Icon Button" have the same identifiers. The second identifier is better to change.

fix-duplicate-id-block


shahednasser:
Awesome thanks!

Issue #918, opened 2020-10-23
[BUTTON IDEA]: Hover Button with revolving item.


harjotkarwal:
**1. Name of button**: Neutron Button

**2. Description**: On hovering, neutron/atomic entity would revolve around on border.

**3. Button Type (Animated, Special, etc...)**: Special

**4. Will you work on it?**: Yes

shahednasser:
@harjotkarwal I'll assign you

harjotkarwal:
#927

harjotkarwal:
#942

harjotkarwal:
@shahednasser Here's my updated code #942

PR #942, opened 2020-10-25
Atom button

harjotkarwal:
Added animated button.
 Referred to issue #918

Issue:

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -160,6 +160,13 @@ $(document).ready(function () {
       classes: "",
       children: [
         {
+          type: "atom",
+          label: "Atom",
+          classes: "atom-btn",
+          shouldHaveRoundedType: false,
+          shouldHaveBlockType: false,
+        },
+        {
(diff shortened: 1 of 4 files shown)


harjotkarwal:
@shahednasser Have a look at my PR.

HeinKhantZaw (approved):

shahednasser (changes requested):
Thank you for your work @harjotkarwal . I think the "atom" that appears on hover should be white in the middle. Also, instead of using Lato please remove it and instead use "Syne" in `asset/css/index.css`. We only use fonts if it's absolutely necessary for the button's style.

harjotkarwal:
> 
> 
> Thank you for your work @harjotkarwal . I think the "atom" that appears on hover should be white in the middle. Also, instead of using Lato please remove it and instead use "Syne" in `asset/css/index.css`. We only use fonts if it's absolutely necessary for the button's style.

Updated.

shahednasser:
Great work thank you!

@all-contributors please add @harjotkarwal for code

PR #950, opened 2020-10-26
Add Vibrate Button

adrsh79086:
Issue:

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -225,11 +225,6 @@ $(document).ready(function () {
           classes: "liquid-btn",
         },
         {
-          type: "vibrate",
-          label: "vibrate",
-          classes: "vibrate-btn",
-        },
-        {
           type: "loading",
           label: "Loading",
           classes: "loading-btn",
@@ -349,6 +344,11 @@ $(document).ready(function () {
           type: "transparent",
           label: "Transparent On Hover",
           classes: "transparent-btn",
+        },
+        {
+          type: "Vibrate",
+          label: "Vibrate",
+          classes: "vibrate-btn",
         },
       ],
     },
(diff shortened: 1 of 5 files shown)


dsnehasish74 (changes requested):
Thank you for your work but can you please add this button in alphabetical order

adrsh79086:
> Thank you for your work but can you please add this button in alphabetical order

okey i will do

assets/js/buttons-examples.js
```diff
@@ -224,6 +224,11 @@ $(document).ready(function () {
           label: "Liquid",
           classes: "liquid-btn",
         },
+        {
```
shahednasser, line 227: As @dsnehasish74 said please place it in the correct alphabetical order. Also, capitalize the first letter in the label

shahednasser (changes requested):
Please change the animation to happen on hover instead.

glebkema:
One more thanks for your PR! 

Please start your include less file with an underscore: `_ vibrate.less`. This helps Gulp distinguish the include files from the main file.

shahednasser:
Also as @glebkema pointed out, please rename the `vibrate.less` to `_vibrate.less` and change the import as well.

src/components/_animated.less
```diff
@@ -25,3 +25,4 @@
 @import "animated/_loading-btn";
 @import "animated/_spring.less";
 @import "animated/_papercurl.less";
+@import "animated/vibrate.less";
```
glebkema, line 28: One more thanks for your PR!

Please start your include less file with an underscore: _vibrate.less. It helps Gulp distinguish the include files from the main file.

assets/js/buttons-examples.js
```diff
@@ -345,6 +345,11 @@ $(document).ready(function () {
           label: "Transparent On Hover",
           classes: "transparent-btn",
         },
+        {
+          type: "Vibrate",
```
dsnehasish74, line 349: please make the first letter of the type small .

dsnehasish74 (commented):
And please make the file name _vibrate.less instead of _Vibrate.less

Issue #883, opened 2020-10-21
[BUTTON IDEA]: Blob Button


shahednasser:
**1. Name of button**: Blob Button

**2. Description**: Button that fills up in a blob style

**3. Button Type (Animated, Special, etc...)**: Animated

**4. Will you work on it?**: No, please comment on this to be assigned

You can see an example of it [here](https://codepen.io/suez/pen/aOgMxy).

wisetok12680:
I would like to work on this..

kavinsingh28:
I am new to this , but i wanna give this a try, can you please elabore more !

mrSidSat:
Can i work on this???

shahednasser:
Yes @mrSidSat I'll assign you

mrSidSat:
Here in this issue i Dont think this example can be implemented without any markup. i tried for a valid solution but this requires use of minimal markup for the blob effect...So i would like to ask if there is any provision to use JS here as it would make this job a lot easy...

shahednasser:
@mrSidSat this project is a CSS only project as of now. I understand your problem but I provided the example just as a reference. You don't have to do it exactly the same. Just try to make use of `:before` and `:after` and do it the way you find best.

mrSidSat:
okay thanks i would try..

Sumit2399:
Hi @shahednasser,I am new to open source contribution.Can you provide me with some more pointer on what has to be done on this issue.Help would be appreciated.

marcoscalia:
Hi @shahednasser, maybe can be helpful this [solution](https://codepen.io/marcoscalia/pen/wvWgRbg)

shahednasser:
@Sumit2399 @marcoscalia someone is working on this issue.

Sumit2399:
Hi @shahednasser,I am trying to add a new button color .Most of the buttons are working fine but some animated buttons are causing issue.Can you help me to understand what could be the problem?

divyanshutewari17:
Can I Work on this?

shahednasser:
I am unassigning @mrSidSat as he is working on a different approach

shahednasser:
@divyanshutewari17 let me know if you still want to work on this

viktoraronfarkas:
@shahednasser I could work on this

UpasanaGhosh:
Can I please take this up if there isn't anyone else working on it? Thanks!

shahednasser:
Sure @UpasanaGhosh I'll assign you

[...]

PR #964, opened 2020-10-27
blob btn added

mrSidSat:
Blob button has been implemented . A colourfull blob appears from eternity to fill the button with letter shining in splendid grace.

Issue: #883


mrSidSat:
error typos

PR #966, opened 2020-10-27
Blinking btn

adrsh79086:
Issue:

### DIFF
--- a/src/components/animated/_Blinking.less
+++ b/src/components/animated/_Blinking.less
@@ -1,0 +1,22 @@
+.blinking-btn&:hover {
+  animation: blinkingText 1.2s infinite;
+  border: none;
+  background-color: none;
+}
+@keyframes blinkingText {
+  0% {
+    color: #000;
+  }
+  49% {
+    color: #000;
+  }
+  60% {
+    color: transparent;
+  }
+  99% {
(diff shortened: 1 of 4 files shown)


dsnehasish74 (commented):
Blink button should be above of bounce button . And for black and white button the animation is not working can you fix that ? Otherwise you did a great work . Thank you .

shahednasser (commented):
In addition to @dsnehasish74 's comments, don't use the black text for all the buttons. If a button originally has a white text keep the white text, and if it has a black text then keep it black.

src/components/animated/_Blinking.less
```diff
@@ -0,0 +1,22 @@
+.blinking-btn&:hover {
```
shahednasser, line 1: Remove the & here it's unnecessary

src/components/animated/_Blinking.less
```diff
@@ -0,0 +1,22 @@
...
+  animation: blinkingText 1.2s infinite;
+  border: none;
```
shahednasser, line 3: remove this line

shahednasser (commented):
Also please rename `src/components/animated/_Blinking.less` to `_blinking.less` without capitalizing the b.

PR #967, opened 2020-10-27
Blob Button Implemented

mrSidSat:
Blob Button is implemented. A colourfull blob falls out of eternity to fill the button and beautifully glaze the text within.

Issue: #883


assets/js/buttons-examples.js
```diff
@@ -833,7 +846,18 @@ $(document).ready(function () {
           totalClasses += ` ${thisButtonClasses}`;
           totalWrittenClasses += ` .${thisButtonClasses}`;
         }
-
+        if(buttonArr[0].type=="blob")
```
shahednasser, line 849: What's the purpose of this code? This button needs to be CSS only so please explain this code.

shahednasser (changes requested):
Thank you for your work I like what you have done! I just have some remarks:

1. The button's concept changed from the "blob button" but I like it. I think we should change the name to Spotlight Button. Make sure to make change to the class name and its file as well. Also, in the label in `assets/js/buttons-examples.js`  just name it "Spotlight" without "btn".
2. I noticed this button only works for one color, but we need it to work for all the button colors in the website, similar to other buttons. So please do that.
3. I've noticed you added some JS code (I added a note about it below) so please explain it as this is a CSS only project.

assets/js/buttons-examples.js
mrSidSat: In the" sButtons " web page the code in buttons-examples.js file automatically  inserts the ".sbtn "class to all the buttons which in turn interferes or overshadows or overwrites the properties written  in  button specific  classes(eg. .blob-btn).
So here I just stopped it from adding the sbtn class to the  div with " blob-btn" class in the page. It works fine for all the rest of the buttons . There is no change in them . This change is blob button specific , would cause no error in the rest of the buttons...
As per the above remarks I would make the necessary changes and introduce a new PR soon..
Thank You.

shahednasser: Thank you for explaining this. However, I think this will cause confusion, as we need the `sbtn` to be the base class throughout the project. You can look at bootstrap for example and how they use `btn` as a base class for their buttons. If there are things in your button that you need to overwrite the properties of the base class `btn`, then try to use `!important` and please remove this code. Thank you.

mrSidSat: Okay...

mrSidSat:
Made a new PR with the spotlight btn

PR #1029, opened 2020-10-30
Changed the color of share-icon to blue

GuilhermeAdias:
Issue: #1029


shahednasser:
Thank you for your work but we don't have this button anymore

Issue #1012, opened 2020-10-30
Enhance sidebar behavior when reaching the footer


shahednasser:
The sidebar currently is hidden behind the footer, we need a better behavior for this to make the design look better.

EnechukwuChibuike:
Can I do this?

shahednasser:
Yes @EnechukwuChibuike I'll assign you

VinayakChopra:
Hi, Can I work on this?

shahednasser:
Yes @VinayakChopra you can

VinayakChopra:
Hi, @shahednasser I have created a pull request with a fix for the given issue, kindly check the same.

TomWBush:
I could try to add media query to Vinayak's solution if he is not still working on this.

VinayakChopra:
@TomWBush I am not working on this now and you may go ahead(@shahednasser I hope you agree too).

PR #1030, opened 2020-10-30
fix sidebar hidden behind footer

AndreJesusBrito:
Fixed sidebar behind footer. Now the footer has a margin of the same size of the sidebar, so it doesn't overlap.

The animation when the footer resizes it’s not perfect, but I believe that is a good improvement.

Issue: Enhance sidebar behavior when reaching the footer #1012

### DIFF
--- a/assets/css/footer.css
+++ b/assets/css/footer.css
@@ -9,6 +9,13 @@ footer {
   position: relative;
   z-index: 12;
   min-height: 20vh;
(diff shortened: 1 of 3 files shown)


shahednasser:
Thank you for your work @AndreJesusBrito but I'm not a big fan of this. I'll add the "hacktoberfest-accepted" label since I appreciate your work. I'll be closing this PR now but if you have another idea you can submit a new PR.

Issue #1013, opened 2020-10-30
JS error on the main page of the site


glebkema:
**1 - Describe the bug**
The browser console shows a Javascript-error on the home page of the site:
https://sbuttons.github.io/sbuttons/index.html
```
Uncaught TypeError: Cannot read property 'left' of undefined
    at S.fn.init.$.fn.isOnScreen (index.js:18)
    at checkSidebar (index.js:71)
    at HTMLDocument.<anonymous> (index.js:153)
    at e (jquery-3.5.1.min.js:2)
    at t (jquery-3.5.1.min.js:2)
```

**2 - Expected behavior of what should have happened (if applicable)**
The error must be fixed as soon as possible, as it stopped working all the scripts on this page.

**3 - Screenshots (if applicable)**
![bug js on mhome-page](https://user-images.githubusercontent.com/3881568/97749954-35156580-1af0-11eb-8737-76d43d78f0bd.png)

**4 - Additional notes**

MSTEWARDSON:
I would like to try and fix this issue.

glebkema:
@MSTEWARDSON OK. I have assigned this issue to you.

MSTEWARDSON:
Update on the issue.

I've narrowed down a few details for this issue. It only happens on the home page. The function call that creates this error comes from the checkSidebar() function within index.js when it calls the isOnScreen() function on the footer element.

Atm Im working into possible solutions to the problem.

MSTEWARDSON:
Alright I have made a PR to fix this issue. Any feedback or suggestions would be appreciated.

Thanks.

PR #1031, opened 2020-10-31
Fixed JS error on the main page of the site #1013

MSTEWARDSON:
Fixed the undefined error that came up for the 'left' value of viewport on the main page. After looking through the code I determined the issue was caused within the function checkSideBar(). Here the html element 'main-head' was used to call the isOnScreen() function. The issue with this and why the error only happened on the homepage was because the homepage does not use the 'main-head' css class.

To fix this issue there is now a check to see if the 'main-head' element exists using .length. 

Issue: JS error on the main page of the site #1013


assets/js/index.js
```diff
@@ -68,12 +68,16 @@ $(document).ready(function () {
...
+    if ($(".main-head").length) {
+      if (!$(".main-head").isOnScreen()) {
+        $(".sidebar").addClass("scrolling");
+      } else {
+        $(".sidebar").removeClass("scrolling");
+      }
     }
 
+
+
```
felops, line 80:
```suggestion
```

MSTEWARDSON:
I have removed the extra space in the checkSidebar() function.

glebkema (commented):
@MSTEWARDSON Thank you for the quick analysis and fixing the problem!
Could you make the same improvement to the footer code below in this same function?

MSTEWARDSON:
Sure I will work on that now.

glebkema (approved):
@MSTEWARDSON Super! Thank you very much. We will be glad if you can help us in other tasks.

@shahednasser I propose to combine this PR.

shahednasser:
Awesome thanks for your work and detailed description!

@all-contributors please add @MSTEWARDSON for code and bug

MSTEWARDSON:
Perfect, thanks!

PR #1032, opened 2020-10-31
style: change animation duration for flash button

### LINKED ISSUE #1019: Make flash button's animation faster
The animation for the flash button is currently slow, it should be faster.

felops:
closes #1019

I think 0.6s is the best option. I made a gif for other speeds I have tested as well:

0.5s
0 5s

0.6s
0 6s

0.8s
0 8s

1s
1s

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -2921,7 +2921,7 @@ a.yellow-btn:active:not(.fill-color-btn) {
 }
 .flash-btn:hover {
   animation-name: flash;
-  animation-duration: 2s;
+  animation-duration: 0.6s;
   animation-timing-function: ease-in-out;
   animation-direction: alternate;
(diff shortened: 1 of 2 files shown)


shahednasser (changes requested):
You've done a great work! Now that the animation is faster it seems laggy due to the animation initially made. Can you please change the `flash` animation by removing the `50%` part as it is not necessary?

felops:
> You've done a great work! Now that the animation is faster it seems laggy due to the animation initially made. Can you please change the `flash` animation by removing the `50%` part as it is not necessar

Sure, I will remove it

felops:
Alright, `50%` removed and PR updated to the last commit.

shahednasser:
Great work thank you!

@all-contributors please add @felops for code

Issue #1017, opened 2020-10-30
Paper curl button's background should be transparent on curl


shahednasser:
In Paper curl button when you hover the background of the "curl" is white but it should be transparent (you can see it when you switch to dark mode)

ZeroniaFX:
Hi can you assign this to me please? Thanks!

Mani-07:
@shahednasser - Can you assign this issue to me?

Mani-07:
@shahednasser  - I have fixed this issue. please check below link.
https://deploy-preview-1046--clever-boyd-7dd108.netlify.app/examples.html#papercurl

PR #1036, opened 2020-10-31
Used dark papercurl when in dark theme.

ZeroniaFX:
Addresses issue #1017. 

Made so it when in dark mode, papercurl uses dark mode's background color instead of white. 
While not transparent, it provides the desired look. 

Before the background was explicitly set to white. I followed this same pattern by doing this. 
Is this workaround okay?

Thank you

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -3112,7 +3112,13 @@ a.yellow-btn:active:not(.fill-color-btn) {
 }
 .papercurl-btn:before {
   background: rgba(0, 0, 0, 0)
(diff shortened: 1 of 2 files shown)


dsnehasish74 (changes requested):
It is good . But you are not supposed  to do that . We need a button which has transparent background . Because this button can also be used in other background color as well the it will not look so good . So rather changing the color can you please make the color transparent so that it looks good in every type of background color . Hope you understand my point

shahednasser:
Great job thanks!

@all-contributors please add @ZeroniaFX for code and bug

PR #1037, opened 2020-10-31
Changed index.html subtitle

drewren25:
I just changed the website's main subtitle to something I think is more concise and gets the point across easier. I also thought saying "Sbuttons" at the end was a little unnecessary so I changed that to the "for your next project". 
This commit says it has 5 commits but 4 of them are just me updating my forked repository. It shouldn't cause any trouble to the master sbutton repo but just wanted to make that clear. Not too sure how to update my forked repo without it showing up as a commit that then has to be merged to the master when I want to make an actual commit so any advice would be helpful, thanks!


shahednasser:
I like your idea thank you!

@all-contributors please add @drewren25 for content

PR #1044, opened 2020-10-31
fix: make loading animations smoother

### LINKED ISSUE #1003: Make loading-btn animation smoother
**1 - Describe the bug**
The loading animation in the some of the loading buttons is not smooth i.e. it snaps back to original position which looks weird.

**2 - Expected behavior of what should have happened (if applicable)**
Simple solution is to just add `alternate` to the loading animations of the buttons that have this problem [here](https://github.com/sButtons/sbuttons/blob/master/src/components/animated/_loading-btn.less).
 
**3 - Screenshots (if applicable)**
captured

karleypetracca:
Issue: #1003

@HADES-01 noted that the animations in select loading buttons is twitchy and does not seem to animate as expected.

This PR makes minor edits to timing and animation settings for all loading button types except the `spin` type with the goal of smoothing out animations.

**Original:**
loading-buttons-current2

**What this PR proposes:**
loading-buttons-proposed2

Resolves #1003

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -2976,7 +2976,7 @@ a.yellow-btn:active:not(.fill-color-btn) {
   margin-top: 10px;
   margin-right: 15px;
   background: #fff;
-  animation: 1.5s bounce linear infinite;
+  animation: 0.7s bounce linear infinite alternate;
   border-radius: 50%;
 }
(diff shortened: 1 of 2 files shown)


shahednasser:
Looks much better thanks!

@all-contributors please add @karleypetracca for code

Issue #1042, opened 2020-10-31
Dark mode text color change 


dsnehasish74:
In dark mode the text under Installation is not clearly visible . I think making it white will be great .
![Screenshot (371)](https://user-images.githubusercontent.com/58779604/97773012-4ef78c80-1b72-11eb-86e0-0dafeb1199f5.png)

shresthalucky:
Fixed on same PR #1045

Issue #1041, opened 2020-10-31
Color Theme is not changing on Home page


dsnehasish74:
**1 - Describe the bug** When you are on home page https://sbuttons.github.io/sbuttons/ the color theme can not be toggle . Though it is working on other pages

shresthalucky:
Fixed with PR #1045

PR #1045, opened 2020-10-31
Fix Color Theme is not changing on Home page

shresthalucky:
Fix Color Theme is not changing on Home page ( issue #1041 )

- Unable to find element with class ```.main-head``` in index page was causing the issue. Resolved it checking length of collection returned by query selector, ```$(".main-head").length```.
- Also added checking condition to footer element, which uses ```$.fn.isOnScreen``` for future prevention.

- Fix Dark mode text color change ( issue #1042 )

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -5,7 +5,6 @@
 
 .installation,
 .cdn .head {
-  color: #413f3f;
   font-weight: 700;
 }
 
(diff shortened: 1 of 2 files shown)


shahednasser:
Awesome thanks!

@all-contributors please add @shresthalucky for code and bug

PR #1046, opened 2020-10-31
[Mani] Fixed issue background color as transparent.

Mani-07:
Fixed issue background color as transparent.

Issue:
https://github.com/sButtons/sbuttons/issues/1017


shahednasser:
Thank you for your work @Mani-07 but someone else is working on this.

PR #1048, opened 2020-10-31
make flash button animation faster

ramkrishn123:
Issue: Flash button animation slow.
Solve: I like to inform maintainers that am changed the animation of the flash button, please check and approve it.


shahednasser:
Thank you for your work @ramkrishn123 but someone else is working on this.

Issue #1016, opened 2020-10-30
Tab's text on hover in dark mode should not change


shahednasser:
In the examples page when it is dark mode, the text on hover becomes white so it's not readable.

felops:
Is this bug when you hover on the tab? I could solve it (:

felops:
Does this bug happens when you hover on the tab? I could fix it (:

eg180:
I'd like to try. Thank you

Mani-07:
@shahednasser - I am interested on this. Can you assign to me?

Mani-07:
@shahednasser - I have fixed issue and raised PR. Can you check and let me know if an concern.
https://github.com/sButtons/sbuttons/pull/1052

shahednasser:
Assigning @felops for commenting first

PR #1065, opened 2020-10-31
style: change tab's text to white when hover in dark mode

felops:
closes https://github.com/sButtons/sbuttons/issues/1016


shahednasser:
Awesome work thanks!

@all-contributors please add @felops for code and bug

Issue #1063, opened 2020-10-31
fix indentation of text


mayankmohak:
**1 - Describe the bug**
![Untitled](https://user-images.githubusercontent.com/63801800/97780392-5b97d700-1baa-11eb-96d3-b40266afab8f.png)

**2 - Expected behavior of what should have happened (if applicable)**

**3 - Screenshots (if applicable)**

**4 - Additional notes**

PR #1067, opened 2020-10-31
add responsive h1 for mobile + delete Quotation mark

airqz:
Issue:#1063

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -236,10 +236,6 @@ body {
   background: none;
 }
 
-.main-head h1  {
-  font-size: 3rem;
-  }
-
 @media only screen and (max-width: 600px) {
   .scroll-top {
     margin-left: 75%;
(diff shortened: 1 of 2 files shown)


assets/css/main.css
```diff
@@ -235,6 +235,11 @@ body {
   color: var(--main-color);
   background: none;
 }
+
+.main-head h1  {
+  font-size: 3rem;
```
shahednasser, line 240: There's already a line that sets the font size in this file for `.main-head h1` so please change that instead of adding a new one

shahednasser (changes requested):

assets/css/main.css
airqz: sorry ...

shahednasser: I see you removed your change and kept the old value. Can you please modify the old value?

shahednasser:
You still haven't fixed the size of `.main-head h1` please do that **by changing the original rule in the file**

shahednasser:
@airqz hi do you have any update on this?

Issue #1020, opened 2020-10-30
Make margin below "extra large" buttons smaller


shahednasser:
The margin below "extra large" buttons is big. It should be smaller. (Change needs to be done in `assets/css/index.css` as this is in the website, not in the button's code)

d3fau7t:
Could I pick this up?

pragapraga:
Shall i pickup ? @shahednasser

eg180:
Good morning. I am interested in helping. Thank you. :)

shahednasser:
@d3fau7t you may work on this as you commented first

d3fau7t:
@shahednasser Does this change actually need to take place in `assets/css/examples.css`?

PR #1074, opened 2020-10-31
Reduced margin below extra-large buttons

d3fau7t:
Reduced margin below extra-large buttons to make visual grouping more obvious.

Issue: https://github.com/sButtons/sbuttons/issues/1020

### DIFF
--- a/assets/css/examples.css
+++ b/assets/css/examples.css
@@ -27,6 +27,10 @@ section {
 }
 
 .button-container .sbtn {
+  margin-bottom: 1em;
+}
+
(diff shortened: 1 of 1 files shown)


shahednasser:
Thank you for your work @d3fau7t but this makes it smaller for all buttons. Can you add a rule that's specific for the extra large button?

d3fau7t:
@shahednasser Just made an update

shahednasser:
Awesome work thanks!

@all-contributors please add @d3fau7t for code

PR #1084, opened 2020-11-01
Link to the example page rather than then the home page from Documentation page and remove target blank because a new tab should not be opened

dsnehasish74:
Issue:

### DIFF
--- a/documentation.html
+++ b/documentation.html
@@ -244,7 +244,7 @@
           </div>
           <div class="text">
             You can find all classes and their corresponding sButtons mentioned in our <a
-              href="https://sbuttons.github.io/sbuttons/examples.html">website</a>.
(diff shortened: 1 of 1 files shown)


documentation.html
```diff
@@ -244,7 +244,7 @@ <h2 class="display-inlineBlock">Basic Button</h2>
           </div>
           <div class="text">
             You can find all classes and their corresponding sButtons mentioned in our <a
-              href="https://sButtons.github.io/sbuttons/">website</a>.
+              href="https://sbuttons.github.io/sbuttons/examples.html">website</a>.
```
shahednasser, line 247: can you change the text from "website" to "Examples page"?

shahednasser (changes requested):

shahednasser:
Great thanks!

Issue #1075, opened 2020-10-31
Inconsistent loading button labels


d3fau7t:
**1 - Describe the bug**
Labels for loading buttons are inconsistent with what the button displays.

**2 - Expected behavior of what should have happened (if applicable)**
Buttons with loading animation on the left side should be labeled "loading left", and those on the right side should be labeled "loading right".

**3 - Screenshots (if applicable)**
![image](https://user-images.githubusercontent.com/57768970/97788994-17e1a500-1b8b-11eb-9769-54fccbb18f39.png)

**4 - Additional notes**
Found in `assets/js/buttons-examples.js`

Wei-J-Huang:
I will try to fix this.

shahednasser:
Hello @d3fau7t thank you for noticing this. Actually the change need to be made in `assets/js/buttons-examples.js`. Let me know if you would like to work on this.

jlieow:
Hi can I attempt to fix the bug?

shahednasser:
Sure @jlieow go ahead

PR #1098, opened 2020-11-02
Fix for Inconsistent loading button labels #1075

jlieow:
Inside "buttons-examples.js", updated text for Loading which is a child of Animated Buttons.

Button animation with loading right 

Issue: Fix for Inconsistent loading button labels #1075

Fixed labels for buttons with loading animation. Button with loading animation on left side are labeled "loading left", and those on the right side are labeled "loading right".

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -240,24 +240,24 @@ $(document).ready(function () {
               classes: "spin right green-btn",
             },
             {
-              text: "loading right",
-              classes: "bounce right red-btn",
+              text: "loading left",
+              classes: "bounce left red-btn",
             },
             {
               text: "loading right",
               classes: "bounce right pink-btn",
             },
             {
-              text: "loading right",
-              classes: "heart-beat right blue-btn",
+              text: "loading left",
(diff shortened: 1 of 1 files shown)


shahednasser (changes requested):
All looks great but I want you to do something additional for me: I think it will look better if the left buttons where together and the right buttons where after. You just need to rearrange them.

jlieow:
> All looks great but I want you to do something additional for me: I think it will look better if the left buttons where together and the right buttons where after. You just need to rearrange them.

Sure. I will keep the colors as they are in their current positions and make the buttons on the top left buttons and buttons on the bottom right buttons.

shahednasser (changes requested):
Great but now there are duplicate buttons so can you please remove all duplicate buttons?

jlieow:
My mistake, I didn't realise.

jlieow:
I have removed duplicate buttons.

I have also rearranged buttons, it looks better with left buttons on left half and right buttons on right half.

shahednasser:
Awesome work thank you!

@all-contributors please add @jlieow for code

Issue #1072, opened 2020-10-31
Hatch buttons perspective bug


AndreJesusBrito:
**1 - Describe the bug**
When the icon of an hatch button gets active the icon does not return to default position on hover out

Steps:
1- hover in a hatch button
2- click it
3 - hover out

**2 - Expected behavior of what should have happened (if applicable)**
On hover out is expected the icon to return to default position

**3 - Screenshots (if applicable)**
![bug](https://user-images.githubusercontent.com/35274086/97784638-ee7b4600-1b97-11eb-9aa2-fcff9a8e2b96.jpg)

shahednasser:
Oh nice catch thanks! Would you like to be assigned to this?

AndreJesusBrito:
@shahednasser sure.

PR #1100, opened 2020-11-03
fix hatch btn perspective #1072

AndreJesusBrito:
Fixed bug on the hatch buttons. Before the icon perspective was not following the perspective of the button when on focus.

Issue: Hatch buttons perspective bug #1072


shahednasser:
Great work thank you!

@all-contributors please add @AndreJesusBrito for code and bug

Issue #1014, opened 2020-10-30
Add support for Dark Mode for buttons


shahednasser:
We need to add support for dark mode, as a lot of the buttons need better support for dark mode, not just for our website, but for the users as well.

This issue will be divided between contributors where each contributor can take up a button to add dark mode support to it. 

Adding the dark mode support need to be done in the following way:

1. Adding support when the parent is `[data-theme="dark"]`
2. Adding support when the button has the class `dark-mode`

So, for each button when adding the support for dark mode, you need to apply it to the above 2. For help on it you can refer to [hatch button's code](https://github.com/sButtons/sbuttons/blob/master/src/components/icon/_hatch.less) as it adds support for them (except adding the `dark-mode` class)

The following buttons need support. Please comment below to take up a button:

- [x] Dashed Button (White and Black Colors) **Taken by @singhtalwinder**
- [x] Glow Button **Taken by @singhtalwinder**
- [x] Neon Border **Taken by @chiragkava**
- [x] Outline (Black and White Colors) **Taken by @shahednasser**
- [x] Stack (Remove white color's border) **Taken by @shahednasser**
- [x] Atom (Yellow, Black and White Colors) **Taken by @shahednasser**
- [x] Diagonal Background (Black and White Colors) **Taken by @TomWBush**
- [x] Expanding Border **Taken by @TomWBush**
- [x] Fill (All types) (Black and White Colors) **Taken by @TomWBush**
- [x] Ripple (Black and White Colors) **Taken by @TomWBush**
- [x] Snake Border **Taken by @TomWBush**
- [x] Transparent on Hover (Black Color) **Taken by @modaserMoj**
- [x] Base Icon (When the icon is with a transparent background, make it white) **Taken By @katyabakun**
- [x] Star Icon (similar to above) **Taken by @TomWBush**
- [x] Hatch (It support the media query and data-theme but doesn't have a `dark-mode` class so that needs to be added). **Taken by @farhanfajarr**
- [x] Colored On Hover **Taken by @farhanfajarr**

dotsandchipsandthings:
Hi @shahednasser can I work on this?

singhtalwinder:
Hey, can I do dashed button?:smiley:

theRoushan:
Can I Fix this Issue?

shahednasser:
Sure @singhtalwinder !

shahednasser:
@theRoushan of course but can you pick a button to work on?

muskuu:
[...]

PR #1116, opened 2020-11-04
Added dark mode support for dashed button.

singhtalwinder:
Changed black button to white button and text color of white button
from black to white in dark mode.

Issue:  #1014

### DIFF
--- a/src/components/styles/_dashed.less
+++ b/src/components/styles/_dashed.less
@@ -35,10 +35,28 @@
 
   &.black-btn {
     .borderRule(@black);
+
+    [data-theme="dark"] &,
+    &.dark-mode {
+      .borderRule(@white);
+    }
+
+    @media (prefers-color-scheme: dark) {
+      .borderRule(@white);
+    }
   }
 
   &.white-btn {
     border: dashed 2px @whiteBorder;
     color: @darkText !important;
+
+    [data-theme="dark"] &,
+    &.dark-mode {
+      .borderRule(@white);
+    }
+
+    @media (prefers-color-scheme: dark) {
+      .borderRule(@white);
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


shahednasser (changes requested):
Please run `npm run compile` for changes to be reflected in `dist/sbuttons.css`

singhtalwinder:
Done

src/components/styles/_dashed.less
```diff
@@ -35,10 +35,34 @@
...
+      .borderRule(@white);
+    }
   }
 
   &.white-btn {
     border: dashed 2px @whiteBorder;
     color: @darkText !important;
+
+    [data-theme="dark"] & {
+      color: @white !important;
```
shahednasser, line 57: Can you please put this line in a new mixin and call it from each? That way if we need to make any changes it will all be in one place. You can put the mixin inside the button.

Also, instead of putting rules separately for `[data-theme="dark"] &` and `&.dark-mode` I think you can put them in the same block of rules. For example:
```
[data-theme="dark"] &,
&.dark-mode {
...
}
```

Please do that in both button colors as it reduces repeated code

singhtalwinder: ok

singhtalwinder: Instead of creating a new mixin, can I use the **borderRule** mixing with white color, like this:

```less
&.white-btn {
    border: dashed 2px @whiteBorder;
    color: @darkText !important;

    [data-theme="dark"] &,
    &.dark-mode {
      .borderRule(@white);
    }

    @media (prefers-color-scheme: dark) {
      .borderRule(@white);
    }
  }
```

shahednasser: That's fine as well

shahednasser:
Great work thank you! You can take other buttons for dark mode as well.

@all-contributors please add @singhtalwinder for code

PR #1127, opened 2020-11-06
Colored on hover

mrSidSat:
### **_Colored-on-hover button added_** 

Issue: #416


src/components/animated/_colored-on-hover.less
```diff
@@ -0,0 +1,108 @@
...
+	0% {
+		color: black;
+	}
+	100% {
+		color: cornsilk;
+		text-shadow: 0px 0px 20px white;
+		box-shadow: inset 0px 0px 5px cornsilk;
+	}
+}
+@keyframes fonntrev {
```
shahednasser, line 11: Name the animations a proper name something that reflects what the animation does. Also, prefix the name with `coloredOnHover` to not cause any problems with other animations.

src/components/animated/_colored-on-hover.less
```diff
@@ -0,0 +1,108 @@
...
+	background-color: transparent;
+	cursor: pointer;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	outline: none;
+	&:hover {
+		animation: fonnt 0.25s ease-in forwards;
+	}
+	.bluebtn{
```
shahednasser, line 35: add `&` before all the button colors. Also, you have this button color named wrong. It should be `blue-btn`

src/components/animated/_colored-on-hover.less
```diff
@@ -0,0 +1,108 @@
...
+		&:hover {
+			animation: fonnt 0.25s ease-in forwards;
+			background-color: blue;
+		}
+		border: 3px double blue;
```
shahednasser, line 40: I don't think using an animation is necessary. Just change the background color on hover and if necessary the color of the button as well, and to make the transition seamless add a transition rule with a duration.

In addition I see you have two borders for the buttons but we don't want that. Please remove the additional button. The button should look exactly like the outline button at first, just on hover the background color and the color of the text changes. So, please refer to the code for outline button for help.

Also, the block code inside the color buttons is the same except for the colors used so create a mixin with this block of code that takes a color as a parameter and call it from each of the buttons. Also make sure to use the project's color variables instead of literal colors. For example instead of `blue` use `@blue`.

shahednasser (changes requested):
Thank you @mrSidSat please run `npm run compile` after you're done making the changes. Also please the style and colors of the buttons should be similar to that of the website. If you have any questions let me know.

src/components/animated/_colored-on-hover.less
mrSidSat: Okay I would make the necessary changes soon

mrSidSat:
Please review the changes..

PR #1130, opened 2020-11-07
Fix #1018. Improve the "Next" and "Previous" buttons

### LINKED ISSUE #1018: Weird behavior in blue "next button"
In the Next Button color blue, there's weird behavior on hover. This only happens in the blue color of this button

sButtons-Simple-and-Creative-Buttons (3)

glebkema:
Fix #1018

1. Currently the arrow on the button appears using the `rotateY` transformation. Perhaps this is precisely why a strange vertical line appeared on the screenshot [in the issue description](https://github.com/sButtons/sbuttons/issues/1018). I've used the `opacity` property instead. It has fewer pitfalls.

2. Noticed that the edge of the word on the button blinks on hover. This is because the pseudo-element with an arrow first covers a part of the word, and then moves to the side. To fix this, I placed the pseudo element outside of the button. There it is invisible thanks to the `overflow` property of the button. Now on hover the button becomes wider and the arrow drives inward.

next-hover

3. Removed a large piece of code regarding the background, color and shape of the pseudo-elements. The background and shape may simply be absent, and only the color of the arrow is important. But it is enough to use the `inherit` value to get the actual color of the button.

4. There is no need for special code for the rounded button too.

5. Added two constants for the indentation of these buttons.

6. Updated the code for the `Previous` button in the same way.

What needs to be improved? What have I not taken into account?

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -18,6 +18,8 @@
 /* END VARIABLES IN special/_toggle.less */
 /* BEGIN VARIABLES IN special/_gradient.less */
 /* END VARIABLES IN special/_gradient.less */
+/* BEGIN VARIABLES IN special/_arrow.less */
+/* END VARIABLES IN special/_arrow.less */
 .sbtn {
   padding: 10px;
   text-align: center;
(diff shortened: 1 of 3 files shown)


shahednasser:
Great work as always thanks!

PR #1131, opened 2020-11-07
Subtitle change

drewren25:
Last time I changed the subtitle of the website to "Design and customize buttons easily for your next project." but I only changed it in index.html! So with this commit I just changed the subtitles in documentation.html and examples.html to match the one in index.html.


shahednasser:
Great thanks!

PR #1132, opened 2020-11-07
Colored on hover

mrSidSat:
Colored on hover button with required changes

Issue: #416


dsnehasish74 (changes requested):
Great work but one little change required the basic and the block button having same size that should not be happening.

src/components/_animated.less
```diff
@@ -25,7 +25,10 @@
 @import "animated/_loading-btn";
 @import "animated/_spring.less";
 @import "animated/_papercurl.less";
+@import "animated/_colored-on-hover.less"
```
shahednasser, line 28: Add ; in the end of the file

shahednasser (changes requested):
@mrSidSat your work is looking better! I'm glad to see your progress. I still have two comments:

1. The button colors are still not the same as the rest of the website colors so please fix that.
2. Please fix the formatting in the file `src/components/animated/_colored-on-hover.less`
3. Apply @dsnehasish74 changes as well

Issue #1022, opened 2020-10-30
Add animated class to animated buttons


shahednasser:
Most animated button's animation only happens on hover, we need to add an `animated` class that triggers the animation without hovering over the button. And when the animation class is added make sure that the animation is not also triggered on hover. In order to avoid repeated code in the buttons, please move the animation or transition to a mixin and call it where it is needed.

Also, some button's animation needs to be continuous, others don't. I have indicated below which is continuous and which isn't 

This issue is going to be split between contributors. Each contributor can take up a button from the list below. Please comment before taking up a button, choosing which between you will be working on.

The following buttons need the `animated` class:
- [x] Atom (continuous) **Taken by @karleypetracca**
- [x] Aura (continuous) **Taken by @TomWBush**
- [x] Bounce (continuous) **Taken by @TomWBush**
- [x] Diagonal Background on Hover (not continuous) **Taken by @TomWBush**
- [x] Expanding Border (not continuous) **Taken by @TomWBush**
- [x] Fill Buttons All Types (not continuous) **Taken by @TomWBush**
- [x] Flash (continuous) **Taken by @TanmayJain17**
- [x] Hinge (not continuous) **Taken by @kalimah-apps**
- [x] Invert (not continuous) **Taken by @karishmapatel27**
- [x] Liquid (not continuous) **Taken by @A3vk**
- [x] Neon (not continuous) **Taken by @kyawkyawsoezhu**
- [x] Next (not continuous) **Taken by @kyawkyawsoezhu**
- [x] Nudge (continuous) **Taken by @modaserMoj**
- [x] Paper Curl (not continuous) **taken by @TomWBush**
- [x] Previous (not continuous) **taken by @TomWBush**
- [x] Pudding (not continuous) **Taken by @AbdulMAbdi**
- [x] Pulse (not continuous) **Taken by @AbdulMAbdi**
- [x] Rainbow (continuous) **Taken by @A3vk**
- [x] Ripple (not continuous) **Taken by @abid1998**
- [x] Scale (not continuous) **Taken by @abid1998**
- [x] Shake (continuous) **Taken By @kyawkyawsoezhu**
- [x] Shine (continuous) **Taken by @kyawkyawsoezhu**
- [x] Skew (not continuous) **Taken by @Hassan-Ib**
- [x] Snake Border (not continuous) **Taken by @rad10wave**
- [x] Splash (continuous) **Taken by @modaserMoj**
[...]

PR #1135, opened 2020-11-10
Add animated class to aura and bounce

TomWBush:
Issue: #1022 

Added animated class to aurapulse and bounce. I only changed less/css files and not the js file so you have to manually add the class on the example page.
chrome-capture


shahednasser:
Awesome work thank you! You can take other animated buttons if you want.

@all-contributors please add @TomWBush for code

PR #1141, opened 2020-11-11
Added dark mode support for glow button.

iamtalwinder:
Changed black button to white button and text color of all buttons
from black to white in dark mode.

Issue: #1014


singhtalwinder:
### The text color of the yellow button stays black like in the dashed button.

#### I think this is where the problem is:
https://github.com/sButtons/sbuttons/blob/9cac44dc7961bbef7421ab1c86b8489643d3d3b0/src/components/_basic.less#L69-L76

#### Adding ```:not(.glow-btn)```  fixes the issue

``` less
&:not(.fill-color-btn):not(.ripple-btn):not(.glow-btn):not(.outline-btn):not(.transparent-btn):not(.diag-btn):not(.neu-btn),
  &.transparent-btn:not(:hover) {
    color: @darkText !important;
  }
```

shahednasser:
Yes @singhtalwinder I agree. Make that change but make sure it doesn't interfere with its style when it's not in dark mode

shahednasser:
Thank you for your work!

@all-contributors please add @iamtalwinder for code

PR #1144, opened 2020-11-13
Add Dark Mode Support for Diagonal Background, Fill, and Ripple

TomWBush:
Issue: #1014 

Please see the demo below. The solution I found for black-btn is to add a grey shadow so that it stands out a bit more than just plain black color. Please let me know if you have a better idea.
diag
fill
ripple

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -1952,13 +1952,11 @@ a.yellow-btn:active:not(.fill-color-btn) {
 }
 [data-theme="dark"] .diag-btn.black-btn.black-btn,
 .diag-btn.black-btn.black-btn.dark-mode {
-  color: #000 !important;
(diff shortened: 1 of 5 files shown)


shahednasser (changes requested):
Looks great but I think for black button it would be best in dark mode if the text color was white when the background is transparent. What do you think?

Also, I just noticed a pre-existing problem in full buttons, which is on hover and `.animated` the height and width is being changed, same for the border which is causing the button to be a little jumpy. Can you remove that? In particular in src/components/animated/_fill.less lines 104, 105, 200, 201, 202, 203

TomWBush:
I have made both of your suggested changes. Please see the new demo.
diag (1)
fill (1)
ripple (1)

shahednasser:
Awesome work as always! Thank you @TomWBush

PR #1151, opened 2020-11-18
added dark mode to black color for transparent on hover 

modaserMoj:
Added dark mode support for Transparent on Hover (Black Color) on issue #1014

### DIFF
--- a/src/components/animated/_transparent.less
+++ b/src/components/animated/_transparent.less
@@ -41,6 +41,11 @@
 
     &.black-btn {
       .onHover(@black, @black);
+
+      [data-theme="dark"] &,
+      &.dark-mode {
+        .onHover(@white, @white);
+      }
     }
 
     &.white-btn {


shahednasser (changes requested):
Thank you for your work @modaserMoj but you still need to run `npm run compile` for the changes to take effect

modaserMoj:
> 
> 
> Thank you for your work @modaserMoj but you still need to run `npm run compile` for the changes to take effect

I seem to run into a problem while running npm start.
I first get the error:
Couldn't open browser (if you are using BrowserSync in a headless environment, you might want to set the open option to false) 
I changed open to false from "local" in node_modules but after that when running npm start it gets stuck at 
] [Browsersync] Serving files from: ./
[1] [Browsersync] Watching files...

How can I fix it?

shahednasser:
@modaserMoj you don't need to run `npm start` you can just run `npm run compile`

PR #1152, opened 2020-11-18
Add darkmode support for colored on hover button + add dark-mode class to hatch button

farhanfajarr:
Changed the text color to white when dark mode (black and white button)
Issue: #1014

colored on hover


shahednasser:
Thank you for your work! You can take other buttons as well if you want, but if you do please comment on the issue before to claim it.

@all-contributors please add @farhanfajarr for code

PR #1155, opened 2020-11-18
Fixing buttons that have more than 3 classes

Hanaffi:
Issue:
Buttons that had 3 or more classes weren't displayed correctly.
Each button should be preceded with a `.` but buttons that have more than 3 classes only has a `.` before the first three class names.

### DIFF
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -808,9 +808,9 @@ $(document).ready(function () {
         }
         // add normal button grid for button
         let newText = (totalWrittenClasses.split(" "));
-        if(newText.length > 3)
+        if(newText.length >=4)
         {
-          for(let i=2;i<newText.length;i++)
+          for(let i=3;i<newText.length;i++)
           {
             newText[i]= '.'+newText[i];
 
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -807,16 +807,13 @@ $(document).ready(function () {
           totalWrittenClasses += ` .${thisButtonClasses}`;
         }
         // add normal button grid for button
-        let newText = totalWrittenClasses.split(" ").filter(x => x.length > 0);
-        if(newText.length >=4)
-        {
+         let newText = totalWrittenClasses.split(" ").filter(x => x.length > 0);
           for(let i=0;i<newText.length;i++)
           {
-            if(newText[i][0]!='.')
-               newText[i]= '.'+newText[i];
+              if(newText[i][0]!='.')
+                newText[i]= '.'+newText[i];
           }
-          totalWrittenClasses =newText.join(" ");
-        }
+        totalWrittenClasses =newText.join(" ");
         normalButtonsGrid.append(
           getButtonHtml(
             totalClasses,
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -807,6 +807,16 @@ $(document).ready(function () {
           totalWrittenClasses += ` .${thisButtonClasses}`;
         }
         // add normal button grid for button
+        let newText = (totalWrittenClasses.split(" "));
+        if(newText.length > 3)
+        {
+          for(let i=2;i<newText.length;i++)
+          {
+            newText[i]= '.'+newText[i];
+
+          }
+          totalWrittenClasses =newText.join(" ");
+        }
         normalButtonsGrid.append(
           getButtonHtml(
             totalClasses,
--- a/assets/js/buttons-examples.js
+++ b/assets/js/buttons-examples.js
@@ -804,16 +804,9 @@ $(document).ready(function () {
           totalWrittenClasses = `.${defaultClass} .${button.classes}`;
         if (thisButtonClasses.length) {
           totalClasses += ` ${thisButtonClasses}`;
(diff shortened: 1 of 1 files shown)


TomWBush:
You are adding an extra dot for the third class, and you are spliting on a single whitespace, which caused problem for red loading-right button. I think regex is a better way to go about it. Also I think your fix is not particularly efficient, because you are constructing the incorrect class names on line 806-807, break it apart on 810, and adding it back up . Why not just add the correct class names on 806-807 to avoid the extra works?

Hanaffi:
@TomWBush I have fixed these two problems.
This should be working very well now.
What do you mean by "Why not just add the correct class names on 806-807 to avoid the extra works?" ?

Hanaffi:
@TomWBush I have fixed these two problems.
This should be working very well now.
What do you mean by "Why not just add the correct class names on 806-807 to avoid the extra works?" ?

TomWBush:
I mean the bug is caused by the 806-807 right? What you did is still having line 806-807 construct the incorrect class name(e.g. ".sbtn .icon-btn .icon-left green-btn"), and then break them into single class name([".sbtn", ".icon-btn", ...]), and finally reconstruct them into the correct class name string(".sbtn .icon-btn .icon-left .green-btn"). So if you can just change 806-807 and let it adds the dot correctly. You wouldn't have to break them down and reconstruct it right?

Hanaffi:
@TomWBush  Yeah , But It will take a lot of string manipulation.
Isn't my solution better/easier?

TomWBush:
It will be a very simple fix if you use regex(probably just 5 lines instead of 10). To be honest, I think your solution makes it harder.

Hanaffi:
@TomWBush  Well , I hate Regex to be honest :D
Here it became only 6 lines.

shahednasser:
Thank you for working on this @TomWBush and @Hanaffi but as I mentioned in the issue someone else is assigned to it so I'll be closing this PR.

shahednasser (changes requested):
Thank you for your work @Hanaffi, however your solution is a hack rather than a bug fix. In a lot of cases where the bug might require a work around or can't be fixed directly, however the solution here can be fixed in a very simple manner on line 807. 

@TomWBush mentioned using regex which is a good idea. However, it can be even simpler than regex, you can use the `replace` function to replace spaces with ` .` (space followed by dot). 

Can you please make the necessary changes?

Hanaffi:
@shahednasser Theres not additional spaces to replace. What do you mean?

assets/js/buttons-examples.js
```diff
@@ -807,6 +807,13 @@ $(document).ready(function () {
           totalWrittenClasses += ` .${thisButtonClasses}`;
```
shahednasser, line 807: If you look at this line @Hanaffi you will see that the problem is here. We're just adding a `.` to `thisButtonClasses`. However, `thisButtonClasses` can have many classes in it. For example `spin left purple-btn`. 

In the example given, `totalWrittenClasses` expected value should be `.spin .left .purple-btn` but in our code it will be `.spin left purple-btn` which means only spin has a `.` before.

What I'm saying this line needs to be changed. In `thisButtonClasses` spaces should be replaced with ` .` which is a space followed by a `.` and a `.` need to be added in the beginning of it. This way `spin left purple-btn` will become `.spin .left .purple-btn`.

Hanaffi:
@shahednasser Done. Can you check it?

assets/js/buttons-examples.js
```diff
@@ -804,7 +804,7 @@ $(document).ready(function () {
           totalWrittenClasses = `.${defaultClass} .${button.classes}`;
         if (thisButtonClasses.length) {
           totalClasses += ` ${thisButtonClasses}`;
-          totalWrittenClasses += ` .${thisButtonClasses}`;
+          totalWrittenClasses += ` .${thisButtonClasses.replace(/\s\s/g," ").replace(/\s/g , " .")}`;
```
shahednasser, line 807: it can be just `.replace(" ", " .")` no need for two replace functions

shahednasser (changes requested):

assets/js/buttons-examples.js
TomWBush: One thing we need to be careful is when classes are seperated by two or more spaces(note line 262, loading right red-btn), which is why I said using regex. Or we can just make it standard that all classes are seperated by one space.

Hanaffi: @shahednasser your suggestion doesnt work. as there are double spaces exist.

shahednasser: @TomWBush thank you for pointing this out as I forgot to notice that. @Hanaffi then you can do the same solution but instead of the first parameter being `" "` it should be a regex. It still stands that there's no need for two `replace` calls

Hanaffi: @shahednasser I tried `.replace(/\s\s+/g, '  .')` 
but it didnt work.

shahednasser: Your regex indicates that there should be at least two spaces. It should be `/\s+/g` instead.

Hanaffi:
Got it. Thanks @shahednasser  🥇 :)

shahednasser:
Thank you for your work!

@all-contributors please add @Hanaffi for code and bug

PR #1174, opened 2020-11-25
Add animated to the Liquid and Transparent on hover

sjohnston92:
Issue: Put in Animated. Feel free to reach out if there's any issues

### DIFF
--- a/src/components/animated/_liquid.less
+++ b/src/components/animated/_liquid.less
@@ -1,40 +1,6 @@
 
 .liquidAnimation(){
-  position: relative;
-  z-index: 0;
-
-  &::before {
-    animation: waterEffect 5s linear infinite;
-    background: rgba(255, 255, 255, 0.5);
-    border-radius: 45%;
-    content: "";
-    height: 400px;
-    left: 50%;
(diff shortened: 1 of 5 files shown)


src/components/animated/_liquid.less
```diff
@@ -1,3 +1,44 @@
+
+.liquidAnimation(){
```
shahednasser, line 2: Why did you put all those rules in the `liquidAnimation` mixin? It should just include the same rule as the hover that was before.

src/components/animated/_transparent.less
```diff
@@ -1,17 +1,25 @@
...
 
 .transparent-btn {
   position: relative;
   transition: 0.3s ease-in-out;
 
-  &:hover {
-    .bc(transparent);
+  &:hover,
+  &.animated {
+    .transparentAnimation();
```
shahednasser, line 21: you should pass `transparent` to this mixin.

shahednasser (changes requested):
Thank you for your work. I have a few comments below.

Also can you please remove all changes in package.json and package-lock.json

PR #1176, opened 2020-11-27
Liquid and transparent

sjohnston92:
Issue: Ended up Fixing those issues I need todo for the liquid and the transparent.

### DIFF
--- a/src/components/animated/_liquid.less
+++ b/src/components/animated/_liquid.less
@@ -1,40 +1,6 @@
 
 .liquidAnimation(){
-  position: relative;
-  z-index: 0;
-
-  &::before {
-    animation: waterEffect 5s linear infinite;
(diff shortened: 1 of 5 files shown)


src/components/animated/_liquid.less
```diff
@@ -38,11 +45,11 @@
     .bc(@color);
 
     &:hover::before {
```
shahednasser, line 47: you need to add `, &.animated:before` as well to work for the animated class. Same for the rule for `:after`

shahednasser (changes requested):
Thank you @sjohnston92 but please remove all changes in `package.json` and `package-lock.json`

As for the `_transparent.less` file, I think you've made unnecessary changes by creating a new mixin and it's not working properly. Instead of creating a new mixin just use the original code and add to `&:hover` rules `, &.animated`.

Issue #1114, opened 2020-11-03
Bug in Atom Button


DhrutikPatel:
**1 - Describe the bug** when we click one atom button animation start but on second click its animation is not stoping.

**2 - Expected behavior of what should have happened (if applicable)** one first click animation should start and on second click it should stop.

shahednasser:
Nice catch. I think this is due because the animation is added on active state as well. Would you like to work on it?

DhrutikPatel:
yes, I will try to fix it.

harjotkarwal:
@DhrutikPatel @shahednasser I think that's not a bug but functionality like the other animated buttons. For refrence, check Aura Pulse button under animated category.

shahednasser:
I'm not seeing the issue in Aura pulse button @harjotkarwal

TomWBush:
The animation stops when you click outisde the button.

shahednasser:
@TomWBush that's true but I think what @DhrutikPatel means is that instead of the animation occuring on `:hover`, it's also occuring on `:active`

shahednasser:
Hello @DhrutikPatel are you still working on this?

TomWBush:
I can work on this

shahednasser:
You can go ahead @TomWBush

PR #1178, opened 2020-12-02
fix bug in atom

TomWBush:
Issue: #1114 

Please see the demo below. Let me know if I misunderstood the expected behavior.
chrome-capture


shahednasser:
Great thank you!

PR #1179, opened 2020-12-02
Add animated to nudge and vibrate

modaserMoj:
Issue:

### DIFF
--- a/src/components/animated/_nudge.less
+++ b/src/components/animated/_nudge.less
@@ -2,11 +2,12 @@
   animation: nudge 0.75s;
 }
 .nudge-btn {
-  &:hover {
+  &:hover,
+  &.animated {
     .nudgeAnimation();
   }
   &.animated {
-    .nudgeAnimation();
+    animation-iteration-count: infinite;
   }
 
   @keyframes nudge {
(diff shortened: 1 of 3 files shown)


src/components/animated/_vibrate.less
```diff
@@ -1,24 +1,32 @@
...
-  }
-  40% {
-    transform: translate(-2px, -2px);
+.vibrate-btn {
+  &:hover {
+    .vibrateAnimation();
   }
-  60% {
-    transform: translate(2px, 2px);
+  &.animated {
```
shahednasser, line 8: Instead of adding the &.animated separately combine it with `&:hover` above.

src/components/animated/_nudge.less
```diff
@@ -1,6 +1,12 @@
+.nudgeAnimation() {
+  animation: nudge 0.75s;
+}
 .nudge-btn {
   &:hover {
-    animation: nudge 0.75s;
+    .nudgeAnimation();
+  }
+  &.animated {
```
shahednasser, line 8: Combine the `&.animated` with `&:hover` and only add in `&.animated` the following:

```
    animation-iteration-count: infinite;
```
THis is to make the animation continuous

shahednasser (changes requested):

src/components/animated/_nudge.less
modaserMoj: I made the changes.

shahednasser:
@modaserMoj thank you for your work!

PR #1180, opened 2020-12-02
Liquid and transparent

sjohnston92:
Issue:


TomWBush:
Have you tested it? It's not working in the preview above.

src/components/animated/_liquid.less
```diff
@@ -37,12 +44,12 @@
     color: #000 !important;
     .bc(@color);
 
-    &:hover::before {
-      top: -120px;
+     &.animated:before {
```
shahednasser, line 47: Why did you remove `&:hover` ? it should be for both. Something like this:

```
&:hover:before, &.animated:before
```
I also think you should combine after with it as well since it's the same rule. It'll be something like this:

```
&:hover:before, &.animated:before, &:hover:after, &.animated:after
```

shahednasser (changes requested):
It seems like you didn't run `npm run compile` and commit the changes so you need to run that for changes to be seen.

I also still need you to remove the changes in `package-lock.json` if you're having trouble with that you can just copy the content of the file from this master repository.

Keep in mind that when you need to make any changes you don't need to create a new PR each time. Just commit the changes and they will be seen here!

shahednasser:
@sjohnston92 I saw you closed this. Will you still be working on these buttons?

sjohnston92:
@shahednasser yes

sjohnston92:
@shahednasser yes still will be working on them, will be submitting the revision later today

PR #1184, opened 2020-12-06
Remove an outdated requirement from Contributing page

TomWBush:
As mentioned in an update in #1014, it is not required to add the dark mode media query to the buttons. So, the corresponding requirement in CONTRIBUTING.md needs to be removed.


shahednasser:
Thank you for noticing that!

@all-contributors please add @TomWBush for documentation

PR #1188, opened 2020-12-08
Dark-mode for neon-border

chiragkava:
Dark mode
dark

Normal
light


shahednasser (changes requested):
Thank you for your work. I know this is an existing bug, but since you're working on it can you make sure that the text of the button is white whenever it is being hovered, with the exception of white button of course

chiragkava:
light new

src/components/styles/_neon-border.less
```diff
@@ -13,6 +13,12 @@
     }
   }
 
+  [data-theme="dark"] &,
+  &.dark-mode,
+  &:hover:not(.white-btn) {
```
shahednasser, line 18: can you also add an exception for `yellow-btn` just so that it will be consistent with other buttons.

chiragkava: just to be clear on hover yellow btn should have black color in day mode? if not pls elaborate

shahednasser: In both modes yellow-btn should have the color black on hover

chiragkava:
captured (2)
captured (3)

src/components/styles/_neon-border.less
```diff
@@ -13,6 +13,12 @@
     }
   }
 
+  [data-theme="dark"] &,
+  &.dark-mode,
+  &:hover:not(.white-btn):not(.yellow-btn) {
```
shahednasser, line 18: add `:not(.white-btn):not(.yellow-btn)` to `.dark-mode` as well.

shahednasser (changes requested):
When you add `dark-mode` class to the button only the color of the text changing without the background color.

src/components/styles/_neon-border.less
chiragkava: how  do you want the background to change? I thought the current background goes well so kept as it is.

shahednasser: Sorry ignore the background part, just fix the one about the yellow and white buttons.

chiragkava: Umm what's the issue with yellow and white they seem to work fine in above gifs.
And  adding :not(.white-btn):not(.yellow-btn) to .dark-mode class doesnt change anything as inside
yellow and white button have explicitly defined color @black on hover.

shahednasser: I have narrowed down the problem. Basically the above line says that when the dark-mode class is added, the color for all buttons should be white. So, we should add to `&.dark-mode` hover as well like `:hover:not(.white-btn):not(.yellow-btn)`

shahednasser: This problem is fixed however buttons like the black button still has  issues, can you test all the buttons please that when you add the class `dark-mode` to them they behave just like when the website is changed to dark mode? Because that one is working fine.

chiragkava: yes there seems to be problem in black button in dark mode its neon glow isn't visible in dark background . changes have been made in it and rest of buttons are working fine.

captured (4)

shahednasser: Ok it's all looking better except for the white button. When adding dark-mode class to white button, it doesn't change behavior like when the dark mode is turned on in the website.

shahednasser:
Great work thank you!

@all-contributors please add @chiragkava for code

Issue #1196, opened 2020-12-10
Toggle Button as a radio button dosent get untoggled in documentation.html file


alok1929:
toggle button gets toggled but cant untoggle it

shahednasser:
Thank you for reporting this! Let me know if you'd like to work on it.

Arnab-Bhakta:
Looks like it's an expected behavior. Radio buttons cannot be untoggled on its own. If you select/toggle other option, then previous one will be unselected/untoggled.

So, the example can be modified like this.
`<input type="radio" class="sbtn toggle-btn" name="selectOption">
    <label>Option1</label>
    <br>
    <input type="radio" class="sbtn toggle-btn" name="selectOption">
    <label>Option2</label>
    <br>
    <input type="radio" class="sbtn toggle-btn" name="selectOption">
    <label>Option3</label>
    <br>`

I would like to make a pr on this.

shahednasser:
Good idea @Arnab-Bhakta . I see you made the changes but still haven't created the PR. Once you're ready please create it and please check out the [contributing guidelines](https://github.com/sButtons/sbuttons/blob/8e8568e858a830a3182102a956ebdbe08284a430/CONTRIBUTING.md). If you need any help let me know.

harshitaaaaaa:
Hey, I want to work on this issue but coudn't find out where to make changes. Do we have to do it in toggle.less? Please guide me with that.

shahednasser:
Hello @harshitaaaaaa unfortunately someone else was assigned to this issue and has resolved it

PR #1197, opened 2020-12-11
#1196 radio button not untoggling fixed

Arnab-Bhakta:
Issue:
#1196 

radio button example modified.


shahednasser:
Looks great but can you either resize the toggle button and make it smaller or position the label base length with the button? just to make it look better. Also instead of "option1" "option2" "option3" add real options like maybe car brands or anything.

Arnab-Bhakta:
What would be approach to resize the buttons?

shahednasser:
@Arnab-Bhakta I think the resizing might require more work, if it's easier just reposition the label

Arnab-Bhakta:
> @Arnab-Bhakta I think the resizing might require more work, if it's easier just reposition the label

Is this fine?
Screenshot (330)
or any other modification?


what should be in the clipboard part? code including the styling or only the core ?

shahednasser:
Yes that looks great. keep the code block as it is.

PR #1201, opened 2020-12-12
Add animated class to Paper Curl and Previous

TomWBush:
Issue: #1022

paper-curl
previous


shahednasser:
Thank you @TomWBush !

PR #1203, opened 2020-12-12
Skew animated class added

Hassan-Ib:
Issue:


shahednasser:
Awesome work thank you!

@all-contributors please add @Hassan-Ib for code

PR #1207, opened 2020-12-15
add hinge, invert, and flash

yuwangcs:
Flash (continuous), Hinge (not continuous) and invert (not continuous) are added and tested. Also, I have run "npm run compile" before I commit. Thanks.

### DIFF
--- a/src/components/animated/_flash.less
+++ b/src/components/animated/_flash.less
@@ -1,22 +1,78 @@
+.flashAnimation() {
+  animation: flash 1.5s infinite;
+}
+
 .flash-btn {
+  position: relative;
   opacity: 1;
 
   &:hover {
-    animation-name: flash;
-    animation-duration: 0.6s;
-    animation-timing-function: ease-in-out;
-    animation-direction: alternate;
-    animation-iteration-count: infinite;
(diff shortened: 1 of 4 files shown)


src/components/animated/_flash.less
```diff
@@ -1,22 +1,78 @@
+.flashAnimation() {
+  animation: flash 1.5s infinite;
```
shahednasser, line 2: I see that you changed the duration, please leave it as 0.6s

src/components/animated/_flash.less
```diff
@@ -1,22 +1,78 @@
...
-    animation-name: flash;
-    animation-duration: 0.6s;
-    animation-timing-function: ease-in-out;
-    animation-direction: alternate;
-    animation-iteration-count: infinite;
-    animation-play-state: running;
+    .flashAnimation();
+  }
+
+  .colorBefore(@color) {
```
shahednasser, line 13: What's this mixen and the rest of the code for? Please clarify

shahednasser (changes requested):
Thank you for your contribution! All looks good, I just have some remarks on the flash button

Issue #1210, opened 2020-12-17
[BUTTON IDEA]: diagonal patterns


shahednasser:
Hey , I have new animated button idea to add : appearing diagonal patterns while hover on buttons.

_Originally posted by @atenasadat in https://github.com/sButtons/sbuttons/discussions/1206_

shahednasser:
@atenasadat you are assigned to this

atenasadat:
Sure. working on it

PR #1212, opened 2020-12-19
Diagonal animation.

atenasadat:
Bevel button  class added. It supports dark mode and also displays in rounded and block.

Issue:
It relates to issue #1210 .


src/components/animated/_bevel.less
```diff
@@ -0,0 +1,118 @@
...
+      border-radius: 20 px;
+    }
+
+    &.block-btn{
+      display: block;
+      width: 100%;
+    }
+  
+    &:link::after,
+    &:visited::after {
```
shahednasser, line 60: Nest the `::after` inside `&:link, &:visited` below

shahednasser (changes requested):
Thank you for your work! I have a few comments:

1. Can you name the button something like "diagonal stripes"? Make sure to change file names and class names as well, and place the button in the correct alphabetical order under "Animated"
2. You have added the code for dark mode in CSS directly. You need to do it in the less file instead and then when compiled, it will be added there automatically
3. Remove all changes in `package.json` and `package-lock.json`
4. Add support for `animated` class. You can learn more about it in issue #1022

Issue #1223, opened 2021-01-05
[BUTTON IDEA]: Striped Button Click Animation


bw-space:
**1. Name of button**: 
Striped Button
**2. Description**: 
We can create a new Animated version of the Striped Button, when its clicked upon, a strip animation can play, like the strips moving diagonally towards the bottom left.
**3. Button Type (Animated, Special, etc...)**:
Animated
**4. Will you work on it?**:
Yes

shahednasser:
Sounds good @bw-space would you like to work on it?

dumbcoder7:
Is someone solving this issue? There is no assignee @shahednasser @bw-space ? I see the `up-for-grabs` label. But the author has mentioned that they are interested to work. Please clarify if possible.

bw-space:
@shahednasser due to some work commitments, I am unable to work on this issue, please assign someone who is interested. Thanks for understanding.

J-16:
Hey, can I work on this?

dumbcoder7:
I wanna work on it too

J-16:
No prob mate, take it :-)

adityaraute:
I have created the button, where the animation plays on hovering.
Here is the animation. 
![captured](https://user-images.githubusercontent.com/43912470/105613408-508ac680-5de8-11eb-8334-5ff57b8c381f.gif)


I shall be raising a PR soon.

PR #1233, opened 2021-01-23
Added Animated Striped Button

adityaraute:
I have added animation to the striped button class, as was requested in #1223 
The animation runs infinitely. 

Here is a demo:
> captured
The button works well in dark mode as well.

Please offer suggestions, if any.


shahednasser:
Sorry for the delay but thank you for the amazing work!

@all-contributors please add @adityaraute for code

Issue #1244, opened 2021-01-28
Change instruction block width


shahednasser:
Make the max width of `.instruction-block` 700px to match the code block.

fleventy-5:
Hi , I'm new to open source  and would like to contribute. I have some experience in HTML, CSS and JS can i do anything to contribute to this  issue

shahednasser:
Sure @fleventy-5 I'll assign you. Let me know if you need any help.

fleventy-5:
Can you tell me the location of the css sheet name  containing ' .instruction-block' class I have clone the project to my local machine but i could not find the css sheet containing  the given class issue

fleventy-5:
I got it the required sheet :) and i have set the max width to 700px what do i have to do after this?

shahednasser:
You should submit a pull request. Please refer to our [contributing guidelines](https://github.com/sButtons/sbuttons/blob/d7a2cb416701d7f1cb751e3b0aa86f6694598df3/CONTRIBUTING.md) to learn more but if you still are not sure how to do it then let me know.

fleventy-5:
Hi there @shahednasser  I have made the required changes and   pull request. I hope i have done it to the required specs.

PR #1249, opened 2021-01-28
Made the max width of .instruction-block 700px to match the code block

fleventy-5:
I have  made  the max width of .instruction-block 700px to match the code block.

Issue:#1244


shahednasser:
Great work thank you!

@all-contributors please add @fleventy-5 for code

PR #1267, opened 2021-01-29
added a new button(.gradient-btn.cyan-btn)

sawantvikram:
I have made this button using CSS. is it ok or should I improve it still???

Issue:


shahednasser (changes requested):
Hello @sawantvikram thank you for your work. We just need to modify the existing box-shadow and make it smaller, we don't need to add a new color. Can you do that?

sawantvikram:
ok i will do that

Issue #1242, opened 2021-01-28
Add padding prior to text that's below buttons in documentation


shahednasser:
In the documentation page when there's a text that's below a button (for example [basic buttons](https://sbuttons.github.io/sbuttons/documentation.html#basic)) padding should be added above the text. This should be done to all same cases throughout the page.

jleckron:
Hi, I would like to attempt to solve this!

jleckron:
I have a fix and a PR ready to go. Please let me know if I need to be officially assigned before I submit PR. (This is my first open-source contribution)

shahednasser:
Hello @jleckron you may submit the PR

PR #1269, opened 2021-01-29
Text padding

jleckron:
Added some text padding to text that appears underneath buttons on the documentation page. 

To do so, I created another CSS class called text-under-button, then changed the classed of the desired texts to text-under-button.  #1242 

Issue: Some text on the documentation page required some padding from the button above it.

### DIFF
--- a/assets/css/documentation.css
+++ b/assets/css/documentation.css
@@ -22,6 +22,14 @@ section {
   font-family: "Syne", sans-serif;
 }
 
+.text-under-button {
+  font-size: 1rem;
+  color: black;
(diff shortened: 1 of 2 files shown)


shahednasser (changes requested):
I like the solution, but instead of changing the class from `text` to `text-under-button` can you just put the two classes for the div, then for the CSS rules for `text-under-button` only put the padding property? That way we don't have to repeat rules in the CSS file. The div can have both classes.

shahednasser:
Great work thank you!

@all-contributors please add @jleckron for code

Issue #1239, opened 2021-01-28
Fix click on download link scroll


shahednasser:
In documentation page, when clicking on the download link "click here" under [download CSS file](https://sbuttons.github.io/sbuttons/documentation.html#downloadcss) link, it scrolls to the beginning of the page. This is due to the fact that link is set to `#`. It should be changed to the id of its container to prevent the scroll.

jleckron:
I would also like to resolve this issue.

shahednasser:
Hello @jleckron I'll assign you

PR #1270, opened 2021-01-29
Download link scroll

jleckron:
To fix the scroll issue, I simply replaced "#" with "#!". I did this instead of replacing it with the id of its container in order to make the page not scroll at all, as long as there is no other element on the page named "!". 

Issue: When clicking the download CSS link on the documentation page, page would scroll to the top. #1239


shahednasser:
Thank you for your work!

@all-contributors please add @jleckron for code and bug

Issue #1260, opened 2021-01-29
Add "Copy" link icon to headings in documentation and examples pages


shahednasser:
In the documentation and examples pages, we need to add a "copy" link icon that copies the link of the current section for the user. 
I prefer a single function to handle both pages on click event for the copy icon.

You can use [this](https://fontawesome.com/icons/link) link icon. It should take the id of the section and add it to the end of the URL as hash (for example, https://sbuttons.github.io/sbuttons/examples.html#gradient)

fleventy-5:
Can I take this one?

shahednasser:
Sure @fleventy-5 go ahead

fluffyP4nd4:
![img](https://user-images.githubusercontent.com/57084379/106310129-21150780-6289-11eb-9344-894ec41951c0.png)
Hey @shahednasser  I was able to add the link icon to all the example buttons. But I have to click the icon twice for the url to be copied. Do you have any solution for this problem??

fleventy-5:
![img.png](https://user-images.githubusercontent.com/78148524/106310597-cfb94800-6289-11eb-8906-2e0a738c1102.png)
Hey @shahednasser  I was able to add the copy link to all the button examples . But I have to click the copy icon twice most of the time for the section url to be copied. Do you have any suggestions for this problem?

PR #1274, opened 2021-01-30
Issue#1260

fleventy-5:
Added link icon to heading in Example and document page. Added a function in index.js to handle link click event in both the pages 
and added some css styling for the link icon in main.css for both the pages

Issue:#1260

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -334,3 +334,12 @@ $(document).ready(function () {
   // Re-determine on/off of the scrollHanlder when the window is resized
   window.addEventListener("resize", addScrollHandler);
 });
+function copyLink(item){
+        var temp = document.createElement('input');
+        text =item.href;
+        temp.value = text;
+        document.body.appendChild(temp);
+        temp.select();
+        document.execCommand('copy'); 
+        document.body.removeChild(temp);
+  }
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -334,6 +334,8 @@ $(document).ready(function () {
   // Re-determine on/off of the scrollHanlder when the window is resized
   window.addEventListener("resize", addScrollHandler);
 });
+
+//function to copy link
 function copyLink(item){
         var temp = document.createElement('input');
         text =item.href;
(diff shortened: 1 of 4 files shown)


AsimAM (approved):

shahednasser (changes requested):
Thank you @fleventy-5 for the great work. I have the following notes:

1. Can you make the icon a little smaller?
2. Instead of using the "onclick" attribute on the HTML elements which will need to be added to every new element in the future, can you assign an event listener to the on click event for `.icon-link` with jquery? Something like this:

```
$("body").on('click', '.icon-link', copyLink);
```

Adding the event on the body first then specifying `icon-link` ensures that any element added to the DOM after load will also have the event listener.

fleventy-5:
@shahednasser  I have made changes  to to create only one event listener for all link click. And also to make link link icon smaller but I need to edit the documentation.css  and example.css since I'm getting this output

image

assets/js/index.js
```diff
@@ -333,15 +333,18 @@ $(document).ready(function () {
...
   // Re-determine on/off of the scrollHanlder when the window is resized
   window.addEventListener("resize", addScrollHandler);
-});
-
-//function to copy link
-function copyLink(item){
+  
+  
+  //function to copy link
+   $(".icon-link").click(function copyLink(){
```
shahednasser, line 339: This will not work for newly added elements on the example page. Please add it like I mentioned. Also remove the naming of function "copyLink" as it is not necessary.

Don't forget to remove all the "onclick"  attributes in both javascript and HTML files

shahednasser (changes requested):

shahednasser:
As for making another PR why not just put the changes in this PR?

assets/js/index.js
fleventy-5:
```suggestion
 $("body").on('click',".icon-link",function (){
```

fleventy-5:
@shahednasser I cant edit the other files. Can you help?

shahednasser:
Why not? You can edit any file and then commit the changes and they will appear in this PR

fleventy-5:
image

The Edit option is deactivated it Grey instead  of black.I only read authorization : ).

shahednasser:
You need to clone the repository and then edit it in your editor, commit the changes and the changes will appear in this PR

fleventy-5:
@shahednasser I'm very sorry for the trouble 
This is the new PR #1281

PR #1284, opened 2021-02-03
documentation issue- icon without bg

pri1311:
Issue: In the documentation page under Icon with no background change the class of the example button from base-icon-btn to icon-btn. This should be changed in the button and the code example


shahednasser:
Great work thank you!

@all-contributors please add @pri1311 for documentation

PR #1285, opened 2021-02-04
Modifycolor

pri1311:
Issue: In the documentation page under Modify Button Colors the link to the variables file goes to sbuttons.less instead. The text should be changed to "_variables.less" and the link as well.


shahednasser:
Great work thanks!

Issue #1240, opened 2021-01-28
Choose a new font


shahednasser:
We are currently using [Syne](https://fonts.google.com/specimen/Syne?query=syne) for the font of the website. Choose a better font for the website in a new pull request

3n1k0:
Hello, can I pick this up?

shahednasser:
Hello @3n1k0 yes you can I will assign you

PR #1290, opened 2021-02-07
Add new font

3n1k0:
Updated font Syne to Montserrat. 
Issue: #1240


shahednasser (changes requested):
Thank you for the great work @3n1k0 . I have noticed that we are using many fonts for different parts of the website (for example Philosopher) so can you remove all additional fonts and just use Montserrat?

shahednasser:
Great thank you!

@all-contributors please add @3n1k0 for code

Issue #1256, opened 2021-01-29
Make header section smaller on small devices


shahednasser:
On small devices (width <= 768px) the header part of the section, i.e. the entire part in the navy blue area, should become smaller. There's a lot of empty space on mobile devices. You can change the height or remove the margin, any solution that you find best fitting

JLambertazzo:
I can work on this!

shahednasser:
Hello @JLambertazzo I will assign you

PR #1297, opened 2021-02-13
Increased Header Responsiveness

JLambertazzo:
Made the header more responsive to screen sizes <768px wide. Slight change in order of media queries to ensure they work as expected.

Issue: #1256


shahednasser (changes requested):
Thank you @JLambertazzo for your work. There is still a responsivity issue you can see it when the width is between 649px and 501px and the height is less than 640px. Can you fix that?

Also please remove any changes in sbuttons.min.css

JLambertazzo:
Fixed the specified issues with responsiveness. I believe the changes to .min.css were automatic actions not any changes I made myself, should I still revert them?

shahednasser:
Great thank you!

@all-contributors please add @JLambertazzo for code

Issue #1263, opened 2021-01-29
Add transition property for Rounded Diagonal buttons


shahednasser:
Transition property with value `all .2s` should be added to the [Rounded Diagonal buttons](https://sbuttons.github.io/sbuttons/examples.html#rounded-diagonal-tl) both top left and top right

angelascode:
I'll resolve this issue also.  Please assign. Thanks

shahednasser:
@angelascode I have assigned you the issue

shahednasser:
Unassigning @angelascode due to inactivity

mainak-debnath:
I can work on this one too

shahednasser:
Sure @mainak-debnath I'll assign you

PR #1309, opened 2021-02-18
Added transition to rounded-diagonal buttons

mainak-debnath:
Added `transition: all 0.2s` to rounded-diagonal button, both top-left and top-right

Issue: #1263


shahednasser:
Great work thanks!

Issue #1311, opened 2021-02-19
Properly center button title


shahednasser:
Button titles are not properly centered, someone used margins instead of properly centering the title.

The title should be properly centered horizontally. You can look below for an example.

![Not centered title](https://user-images.githubusercontent.com/27354907/108503678-773ffe00-72bd-11eb-8e2d-1fb8eeb5fcdc.png)

3n1k0:
Hi @shahednasser, can I pick this up?

shahednasser:
Hello @3n1k0 yes I'll assign you

PR #1318, opened 2021-02-19
Centered button sub-headers

3n1k0:
Centered section sub-headers on the example page. 

Issue: #1311


shahednasser (changes requested):
Thank you for your work but this doesn't fix the issue. The issue is that the headline is not properly centered, it's centered by margining it so that should be fixed

3n1k0:
Hi, I had to add a new container to be able to center the sub-headings with the icons. Let me know if you think it's optimal!

shahednasser:
Great work thanks!

@all-contributors please add @3n1k0 for bug and code

PR #1322, opened 2021-02-22
Fix punctuation typo in the documentation

dafiulh:
Issue: Space should be placed after ".", not before


shahednasser:
Awesome thanks for finding that!

@all-contributors please add @dafiulh for documentation

PR #1331, opened 2021-02-25
Updated hinge animation

kalimah-apps:
Hinge is now triggered by hover and adding animated class.

Issue:
#1022


dsnehasish74:
If I am not wrong you already did a pull request so it is better to do a commit on that rather making a new pull request.  Thankyou

shahednasser:
As @dsnehasish74 said, I'll be closing this PR.

PR #1333, opened 2021-02-26
Improve footer design

Bc-coding:
I've made some changes to the footer layout in the index.html, documentation.html and examples.html. As well as the corresponding styles in main.css and footer.css. Please review the changes. Thank you.

Issue:


shahednasser (changes requested):
Hello thank you for contributing! I have a few comments:

1. Please move the links to the first or second section because it's not visible due to carbon ads placement
2. Remove changes in package-lock.json

PR #1334, opened 2021-02-27
Improve footer design

Bc-coding:
Hello Shahednasser, please see the further changes I made as per your comments. I moved footer links to the 2nd section as well as removed the changes in package-lock.json. please take a look the updated version. let me know what you think. thank you! 

Issue:


shahednasser (changes requested):
Thank you for making the changes but the same problem persists which is that part of the footer will not be visible. Let's make it two columns instead of one

Note: You don't need to open a new PR each time, simply commit changes to the branch you created for this PR

Bc-coding:
Hi, thanks for your feedback! I've made the changes to avoid the content being hidden partially. Please take a look and I appreciate your thoughts.

shahednasser:
That's a nice approach but I still think that having two columns in the footer instead of three would be better. I think placing the text in the last column of the footer should be below the logo and the resize the logo to make it smaller

Bc-coding:
Hi Shahed, Thank you for your reply. I have some questions regarding your comments before I make further changes. I appreciate your reply. 
1. having two columns in the footer instead of three would be better. -- a) do you mean having two columns in the desktop screen? If so, should I place logo in the first column, in the second column I place links  and the desc as two separate nested div?  b) in the small screen, the same concept applies but we have two rows instead.
2. placing the text in the last column of the footer should be below the logo and the resize the logo to make it smaller -- Do you mean the text should be below the logo for the small screen size only? Or do you mean for all screen sizes?

shahednasser:
Hi @Bc-coding , here's a description of what I mean to answer your questions:

In the big screen it should be two columns. The first column should have the logo but smaller than it is now with the description below it and the second column should have the links.
On smaller screens, it should be all below each other. logo first, then description, then links. 
I also suggest adding some padding to the footer (more than it is now).

Let me know if you have any other questions.

Bc-coding:
Hi Shahed,

I've made some changes to the layout into 2 columns. Please take a look and let me know what you think.

documentation.html
```diff
@@ -525,6 +525,8 @@ <h4 class="display-inlineBlock">Adding dark-mode as an attribute</h4><a  href="#
   <span class="scroll-top" data-scroll="up" style="display: none;">
     <i class="fa fa-chevron-circle-up" aria-hidden="true"></i>
   </span>
+<<<<<<< HEAD
```
shahednasser, line 528: Please remove all new lines containing "<<<<<" or ">>>>" or "====" or anything similar as it is the result of issues in merging

shahednasser:
Unfortunately this PR has gotten messy and I'm unable to help you resolve this as well. I will be closing this now but please submit another PR and hopefully we can work through this there. Thank you.

Issue #1346, opened 2021-03-13
Projects Using Sbuttons Page


dsnehasish74:
**1 - Describe the bug**
When we hover a link of the dropdown menu of this projects page Its background turns into white that is looking bit odd rather I think the design should be like home page dropdown links.

**3 - Screenshots (if applicable)**
![Screenshot (545)](https://user-images.githubusercontent.com/58779604/111017270-5c523c80-83d8-11eb-94f5-b5643bcc5e6d.png)

owen-young:
I am currently working on this issue.

PR #1347, opened 2021-03-14
Fix for #1346, changed where css files were imported in projects.html

owen-young:
Issue: #1346

I changed where the links to the local stylesheets occurred in order to match how it works in index.html. There were many rules being erroneously overwritten as seen here:
image

After placing the local stylesheets after the bootstrap/fontawesome ones, the dropdown menu now has the correct highlighting. Along with this, some rules in projects.css were removed so that it has the same applied rules as the dropdown in index.html. The removed rules served no purpose (that I could see, I apologize if these were put in for another reason).

### DIFF
--- a/assets/css/projects.css
+++ b/assets/css/projects.css
@@ -75,11 +75,3 @@ main {
   color: #fff;
 }
 /* dark mode style ends here */
-.btn-group ul {
-  background: var(--sidebar-color);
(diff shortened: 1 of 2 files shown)


shahednasser:
Thank you for your work, but this issue now won't be applied considering the website has changed. again thank you!

PR #1349, opened 2021-03-17
Added animated class to Liquid and Rainbow button

A3vk:
I added the animated class to the Liquid and the Rainbow button. I saw that the hover effect was unique for each color button. I updated this, so this is now only specified one time in the CSS file.

Issue: #1022


shahednasser:
Awesome thank you!

@all-contributors please add @A3vk for code

Issue #1354, opened 2021-03-17
Cannot refresh non-home page


TomWBush:
**1 - Describe the bug**
When refreshing the site on any pages other than the home page, the website gives up with a 404 status.

**2 - Expected behavior of what should have happened (if applicable)**
The website should be able to refresh in any page.

dsnehasish74:
I want to work on this issue. This can be done easily by adding the _redirect file in the public folder

TomWBush:
Sure you can work on it.

shahednasser:
Thank you for catching that @TomWBush !

PR #1356, opened 2021-03-17
Resolve Refresh Problem

dsnehasish74:
issue : #1354
This should resolve this problem


shahednasser:
Great work thanks!

@all-contributors please add @dsnehasish74 for bug

PR #1363, opened 2021-03-22
Drop Down Menu width in small devices

rad10wave:
Increased the size of div to accommodate the texts without line break in small devices


shahednasser:
That's great @rad10wave but as we are using Tailwind CSS, can you use one of its width classes instead? I think it would be a better idea

rad10wave:
@shahednasser please check the latest, i used Tailwindcss for adjusting the width

shahednasser:
Great work thank you!

@all-contributors please add @rad10wave for code

Issue #1378, opened 2021-04-20
Toggle button not fully visible on Firefox


andypants888:
Hi! I'm Andrew and I'm new to contributing and coding in general so please let me know if I have missed some obvious or common detail. Thank you. 

1. Bug description below:

**When using the toggle button feature on Mozilla Firefox, the toggle button is partly hidden along the left/right padding at page-load. Once the user clicks on the toggle button, the full button becomes visible.** 

- Partly hidden - Firefox 87 (desktop)
- Partly hidden - Firefox 88 (android)
- Normal - Brave (android)
- Normal - Chrome (android & desktop)

2. Expected Behavior:

The toggle initial view should be same on both Firefox and Chrome.

3. Screenshots:

Here is a JS Fiddle showing the problem: [LINK](https://jsfiddle.net/uL950jgd/2/)

Overview:
![toggle_bug_screenshot](https://user-images.githubusercontent.com/79945330/115471381-ba025180-a1ec-11eb-9d98-d55f44ed3af0.png)

While inspecting element:
![image](https://user-images.githubusercontent.com/79945330/115471332-a48d2780-a1ec-11eb-924e-533633db0d3c.png)

4. Additional:

I tried setting the padding from 10px to 0px (in the css input), but this caused the white middle button to be incorrectly positioned. 

I am thinking maybe the way to fix this is to look in the .less file, adjust the padding rules for the "input checkbox", adjust the white button animations, recompile, and test in various browsers.

Do you guys have any thoughts? Or opinions on how to approach this?

I'd like to work on this, but not cause extra problems... this would be my first real contribution to any project.

shahednasser:
Hello Andrew. Sorry for the late delay and thank you for your interest in contributing to sButtons!

That was a nice catch. I'll assign you to this issue. Let me know if you need any help while working on this.

andypants888:
Hi shahed! 

No worries at all. I've been working on the issue this week and I've been getting some trouble from the lessc compiler (maybe prettier too)

I planned to make 3 changes to 2 files, in toggle.less and variables.less

When I run npm start, my changes to the .less files are not compiling as expected.  I have tried running npm start before making the changes, but then prettier will loop repeatedly without stopping.

My main problems are:

1. lessc is causing unexpected changes in slides 4 & 6 to the hover in the snake-btn, and to the font-face src version.

2. after running prettier, my git shows that 89/91 files have been modified, but no text inside the files have actually been modified.

[...]

PR #1380, opened 2021-05-07
fixed issue #1378

imkarimkarim:
fixed issue:  Toggle button  not fully visible on Firefox #1378 
sorry for not taking permission to work on this issue. I :eyes:   at the issues and the replies were not fast(at least these days). thanks for your time. happy coding :game_die:

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -2810,7 +2810,7 @@ a.yellow-btn:active:not(.fill-color-btn) {
   right: 0;
 }
 .snake-btn:hover:before,
-.snake-btn.animated:hover:before {
+.snake-btn.animated:before {
   width: 100%;
   height: 100%;
   opacity: 1;
@@ -2821,7 +2821,7 @@ a.yellow-btn:active:not(.fill-color-btn) {
   border-color: #333;
 }
 .snake-btn:hover:after,
-.snake-btn.animated:hover:after {
+.snake-btn.animated:after {
   width: 100%;
   height: 100%;
   opacity: 1;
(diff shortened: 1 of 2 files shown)


andypants888:
Hi karim, no problem. I was wondering what you think of the hover pseudo-class that have been changed on the snake buttons? 

This happened to me too... I don't understand why changing the padding from the toggle.less would change the hover pseudo-state.

Thanks.

imkarimkarim:
hi @andypants888 . it got me wondering too. this should be something with less compiler. i don't know much about less... AND we all are learning :) learning never ends :smile:

shahednasser (changes requested):
Hello @imkarimkarim, thank you for your work. Usually, I don't accept pull requests if the issue is already assigned to someone else, but since @andypants888 is ok with it then it's ok.

Your fix works on firefox for input elements, however, you added `margin-left: 10px;` to all toggle button types (not just input elements) that messed up button elements and a elements. Please remove that.

As for the issue you and @andypants888 brought up regarding the snake button, it's definitely strange as it seems that some pull requests have this issue, whereas others don't. Maybe depends on the OS? I'm not sure, but there are two things to note:

1. The compiler's output should be similar to yours, meaning that the `:hover` shouldn't be there.
2. There's a problem with the left and right borders, as they're not showing. I tested it using the CDN as well, and this problem is occurring there too. But, when using the LESS files there's no bug here. Truly strange.

I guess I'll need to open a new issue with this. As for now, please work on the fixes I mention above. Thank you.

shahednasser:
This issue has been fixed by @andypants888 thank you

Issue #1383, opened 2021-05-10
Change the website to use the CSS in the repository


shahednasser:
At the moment, the website under `/website` uses the NPM package. This means that changes in submitted PRs cannot be previewed and any changes in sButtons can only be shown if deployed on NPM.

We need to make changes in the website so that instead of using the sButtons NPM package, it uses the files in this GitHub repository.

swetharnaik:
Hi @shahednasser Am contributing to open source first time. If you are comfortable with it, can I take this up?

PR #1385, opened 2021-05-11
referencing files in repository

swetharnaik:
Dependency sbuttons references the git repo sButtons/sbuttons now instead of the latest npm package release

Issue: #1383 Change the website to use the CSS in the repository

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -18,7 +18,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
     "react-toastify": "^7.0.3",
-    "sbuttons": "^1.2.1",
+    "sbuttons": "sButtons/sbuttons",
     "use-dark-mode-hook": "^1.1.1",
     "web-vitals": "^1.0.1"
   },


shahednasser:
Great thank you!

@all-contributors please add @swetharnaik for code

shahednasser:
To iterate on the npm link part, we can have a prebuild script in package.json to do that

swetharnaik:
@shahednasser No problem. Thanks

swetharnaik:
@shahednasser So the following is the pre-build script I wrote in package.json-
    _"prebuild": "cd ../ & npm link & cd website & npm link sbuttons",_

But this seems to not work though. After running npm run build, am receiving this error- 
      _ModuleNotFoundError: Module not found: Error: You attempted to import ../../dist/sbuttons.css which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
      You can either move it inside src/, or add a symlink to it from project's node_modules/._

Even though I can see the symlink has been created in website's node_mdoules, it throws this error. Can you give this script a try?

shahednasser:
Hello @swetharnaik sorry for the late reply. I tested this script and it worked fine for me except for the fact that you're using `&` instead of `&&`, what OS are you running on? Also what node version?

shahednasser:
@swetharnaik ok so I was able to find a way to resolve this using `craco`. What we need to do is in `website/craco.config.js`, replace the current `webpack` object with the following:

```
webpack: {
      configure: webpackConfig => {
        const scopePluginIndex = webpackConfig.resolve.plugins.findIndex(
          ({ constructor }) => constructor && constructor.name === 'ModuleScopePlugin'
        );
  
        webpackConfig.resolve.plugins.splice(scopePluginIndex, 1);
        webpackConfig.watch = true;
        return webpackConfig;
      }
    },
```

Then, in `website/src/index.css`, change the import to point directly to the `sbuttons.min.css` from the repository:

```
@import url('../../dist/sbuttons.css');
```

This way we don't need to add any prebuild script

swetharnaik:
@shahednasser No issues for the late reply. It was the holidays. Belated Happy Eid!
Yea this works. The create-react-app inherently doesn't allow us to import from outside the project because of the 'ModuleScopePlugin'.
So even though I had created symlinks using "npm link" it didn't work because of this inbuild restriction that react has. The webpack change in craco.config.js made all the difference. Shall I create a PR to accommodate this change?

P.S I work on Windows.

shahednasser:
Thank you!

Yes, I think this is the best solution for now. Please submit a PR with it.

Issue #1386, opened 2021-05-11
Fix copy button


shahednasser:
In the Examples page, the copy button for the classes works fine, but it's copying the text as is. It should copy the text and replace all `.` with spaces

grimxreaper:
Hi there! Would this be a fix in JS? I've been looking through the code base and see that most of the code is written in Less so I figured I would ask in case I could help out with this if you could point me to the file in which the change would be needed. Thank you @shahednasser !

grimxreaper:
Awesome! To clarify, is the change requested to go from this copied:
`.sbtn.glow-btn.blue-btn`
to this:
`sbtn glow-btn blue-btn` ?

shahednasser:
yes that's what we need. But also you need to be aware that this component is used in different parts of the website, not just in the examples page, so it shouldn't affect its usage elsewhere.

I'll assign you. Let me know if you need any help

grimxreaper:
Ah okay, I will check for that.
I'm trying to test my changes but npm start opens a browser page with the message "Cannot GET/" and this message in the terminal. Have you encountered this error before?
<img width="625" alt="Screen Shot 2021-05-11 at 11 17 44 AM" src="https://user-images.githubusercontent.com/64915527/117841967-7cbd3c80-b24b-11eb-9aec-2f5c2cba1dfc.png">

shahednasser:
You shouldn't be pushing to the upstream. Please refer to [this](https://github.com/sButtons/sbuttons/blob/master/CONTRIBUTING.md#create-pull-request) section in the contributing guidelines

PR #1394, opened 2021-05-11
Bug/copy btn text

grimxreaper:
- [x] read the contrib guide

Removed periods and replaced them white blank spaces when copying the code for a button by adding a function in /Components/Button.js. I tried to make the change in /Components/Code.js as you noted but nothing was altering the text being copied until I found that it was within `getCopyClasses` in Button.js that the periods were being inserted in, and so that is where I made the change.
Please let me know what you'd like me to change though, as I know you said this would impact behavior in other places on the website. I checked the functionality with other button types, and everything behaved normally, so I'm not sure what I should be checking for. Do let me know! Thank you!

Issue: Fix copy button [#1386](https://github.com/sButtons/sbuttons/issues/1386)

### DIFF
--- a/website/src/components/Button.js
+++ b/website/src/components/Button.js
@@ -24,13 +24,18 @@ function Button ({classes, text, htmlContent}) {
 
     const copyClasses = classes ? getCopyClasses(classes) : ""
 
+    function testFunc (codeWithPeriods) {
+        classes.replace('.', " ")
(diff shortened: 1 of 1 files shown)


shahednasser:
Great work thank you!

@all-contributors please add @grimxreaper for code and bug

PR #1441, opened 2021-08-23
added box-shadow property

Anshu-jindal:
I have added a standard property of box-shadow where -moz-box-shadow and -webkit-box-shadow already just for the fallback in case both of them don't work.

Issue: There are some browsers that may not recognize -moz-box-shadow and -webkit-box-shadow. So, it is always safe to have a fallback.

Hope you like it.

Thanks,
Anshu Jindal

### DIFF
--- a/dist/sbuttons.css
+++ b/dist/sbuttons.css
@@ -472,6 +472,7 @@ a.yellow-btn:active:not(.fill-color-btn) {
 .key-btn.pink-btn {
   border-top: 2px solid #d53caf;
   background-color: #d53caf;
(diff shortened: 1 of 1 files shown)


shahednasser:
Hi @Anshu-jindal , thank you for contributing and this is a great idea. However, changes should not be made in `sbuttons.css` or `sbuttons.min.css`. 

Please make changes in the necessary files under `src` then run `npm run compile`. You can read more in the [contributing guidelines](https://github.com/sButtons/sbuttons/blob/271dc93e0e524857fa3a8ee75909a68b216c6194/CONTRIBUTING.md)

PR #1448, opened 2021-08-24
added 3d flip checkbox button

Anshu-jindal:
I have added code for 3d flip checkbox button

Issue: Added a new button in the repository named "3d flip checkbox" under type: "Special"

### DIFF
--- a/src/components/_special.less
+++ b/src/components/_special.less
@@ -7,3 +7,4 @@
 @import "special/_code-btn.less";
 @import "special/_rainbow-border.less";
 @import "special/_disco.less";
+@import "special/_3d-flip-checkbox.less";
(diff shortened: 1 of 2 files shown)


shahednasser:
Thank you @Anshu-jindal for your contribution. Unfortunately, the buttons we implement rely on only using one HTML element (usually `<button>`) with the classes we create. You can use pseudo elements like `:before` and `:after`, but that's it. You can refer to how that is with all the other buttons. 

So, you need to change the button to only need one class and not rely on container elements or children elements.

Anshu-jindal:
Okay, I will update my code and will pull the request after doing the required changes

Anshu-jindal:
I am sorry, but I am unable to convert HTML/CSS code into less code such that one CSS class can perform this feature alone. If anyone can help then maybe I will be able to do this but it is difficult for me to do it alone.

shahednasser:
It's ok @Anshu-jindal . I'll close the PR and the issue. thank you for your work.

PR #1465, opened 2021-08-31
Edited the background color of toggle button to highlight the ON and OFF state

snehanagari:
I changed the background color of toggle button to highlight OFF state. 
Issue:1 - 
The toggle button is shows same color for both ON and OFF state.


shahednasser:
Thank you for your work @snehanagari but that does not have to do with the bug. Also, we do not edit the CSS files directly, we edit the LESS files.

PR #1476, opened 2021-09-14
Changes to heading

aviral14112001:
Created new ID  heading in      (website/src/index.css)
few minor changes in (website/src/components/Hero.js)

### DIFF
--- a/website/src/components/Hero.js
+++ b/website/src/components/Hero.js
@@ -6,7 +6,7 @@ function Hero () {
     return (
         <div className="lg:mb-32 mb-10">
             <div className="container mx-auto px-4">
-                <h1 className="lg:text-7xl lg:mt-16 mt-5 text-gray-600 dark:text-white text-5xl underline-red">
+                <h1 className="lg:text-7xl lg:mt-16 mt-5  dark:text-white text-5xl " id="heading">
                     Simple and Creative buttons with sButtons.
                 </h1>
                 <div className="lg:mt-10 mt-9">
--- a/website/src/components/Hero.js
+++ b/website/src/components/Hero.js
@@ -6,7 +6,7 @@ function Hero () {
     return (
         <div className="lg:mb-32 mb-10">
             <div className="container mx-auto px-4">
-                <h1 className="lg:text-7xl lg:mt-16 mt-5  dark:text-white text-5xl " id="heading">
+                <h1 className="lg:text-7xl lg:mt-16 mt-5  dark:text-white text-5xl bg-clip-text">
                     Simple and Creative buttons with sButtons.
                 </h1>
                 <div className="lg:mt-10 mt-9">
--- a/website/src/components/Hero.js
+++ b/website/src/components/Hero.js
@@ -6,7 +6,7 @@ function Hero () {
     return (
         <div className="lg:mb-32 mb-10">
             <div className="container mx-auto px-4">
-                <h1 className="lg:text-7xl lg:mt-16 mt-5  dark:text-white text-5xl bg-clip-text">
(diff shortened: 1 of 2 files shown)


shahednasser (changes requested):
Hi @aviral14112001 thank you for your work! it looks great, but as we are using Tailwind CSS, it would be best to use the utility classes when possible. 

So, please remove the following from the CSS properties for `#heading`:

```css
background-clip: border-box;
-webkit-background-clip: text;
background-clip: text;
```

and instead add to `h1` the class `bg-clip-text`.

You can leave the rest as is. Let me know if anything is unclear.

aviral14112001:
I just  have to create  a new class  like 

h1 . big-clip-text
{.........
..........
.........}

shahednasser (changes requested):
Thank you for making the changes @aviral14112001 but that's not what I meant.

You added `bg-clip-text`. That's good. Now, you need to remove:

```css
-webkit-background-clip: text;
background-clip: text;
```

from `h1.bg-clip-text`.

After that, add a unique class to the h1 element. For example, `.heading`, and place the CSS properties that were previously for `h1.bg-clip-text` in `.heading`.

Let me know if you have any questions.

website/src/index.css
```diff
@@ -151,3 +151,8 @@ body {
     hsla(203, 11%, 15%, 0.4);
   color: #aaa !important;
 }
+.bg-clip-text {
+  background: -webkit-linear-gradient(#ff8533, #ff4da6);
+  background-clip: border-box;
+  -webkit-text-fill-color: transparent;
+}
```
shahednasser, line 158: Can you change the class these properties are added to to a unique class? We don't want these rules to be applied to all future uses of `.bg-clip-text`

website/src/components/Hero.js
```diff
@@ -6,7 +6,7 @@ function Hero () {
     return (
         <div className="lg:mb-32 mb-10">
             <div className="container mx-auto px-4">
-                <h1 className="lg:text-7xl lg:mt-16 mt-5  dark:text-white text-5xl bg-clip-text">
+                <h1 className="lg:text-7xl lg:mt-16 mt-5  dark:text-white text-5xl heading">
```
shahednasser, line 9: Add `bg-clip-text` here

website/src/index.css
```diff
@@ -151,8 +151,8 @@ body {
     hsla(203, 11%, 15%, 0.4);
   color: #aaa !important;
 }
-.bg-clip-text {
+ .heading {
   background: -webkit-linear-gradient(#ff8533, #ff4da6);
-  background-clip: border-box;
+   -webkit-background-clip: text;
```
shahednasser, line 156: Remove this line

shahednasser (changes requested):

website/src/index.css
aviral14112001: I can't  remove  the webkit-background-clip : text ; 
Line the gradient colour  is in the background if I remove this the text  wouldn't be visible

shahednasser: I've tested it, and using the following:

```css
background: -webkit-linear-gradient(#ff8533, #ff4da6);
-webkit-text-fill-color: transparent;
```

with the class `.bg-clip-text` should work successfully.

You now have the class `.bg-clip-text`. Please update the CSS to the following:

```css
.heading {
  background: -webkit-linear-gradient(#ff8533, #ff4da6);
  -webkit-text-fill-color: transparent;
}
```

shahednasser (changes requested):

shahednasser:
Amazing! Thank you for your work!

@all-contributors please add @aviral14112001 for code and idea

Issue #1446, opened 2021-08-24
Add search bar in the examples and documentation pages


shahednasser:
In the [Examples](https://sbuttons.netlify.app/examples) and [documentation](https://sbuttons.netlify.app/documentation) pages, we should add a search bar that allows us to search through the sidebar items. The search bar should be a reusable component that would allow us to use it in any sidebar.

The search bar doesn't need to do anything fancy. Just search through the sidebar items and hide/show the items based on the search term.

varunwho:
Hey can I try working on this?

gursparsh:
Hi! I would like to try working on
 this issue.

shahednasser:
I'll assign @varunwho for commenting first. Thank you both.

varunwho:
working on it

shahednasser:
Thank you for the update @varunwho

varunwho:
I'm having a doubt as to how do I filter the search bar in this? Normally I know if there's any array we can use the filter method or something, so here can you give some hints or some light as to how do I implement it.
(I saw that you've used Sidebarlink text attribute for that)

shahednasser:
That's true, thank you for pointing it out.

I think this can be done by creating a new state in `Sidebar` with initial value as the prop `children`. Then, when the search input's value changes, you traverse the `children` array, check the value of the `text` prop like this:

```js
children[i].props.text
```

check if the text contains the query entered in the input. Basically, filter out the elements in the children array that do not contain the search query, then set the state you created to the filtered array. You can use the [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) method to make it easier.

Finally, change the rendered children in the sidebar to be the state you created.

Let me know if this is clear or not and if you need any additional help.

varunwho:
I'll try, thanks

shahednasser:
Hello @varunwho any update on this?

varunwho:
I'm trying but actually stuck rn, give me 1-2 days. I'll try to do it

varunwho:
@shahednasser  I've done almost 90% of the work, just stuck on the remaining 10%. will try to do it tomorrow itself

varunwho:
@shahednasser the search filter thing is working perfectly on the docs page but the Examples page has a different logic in rendering the sidebar I think so, I'll have to see to it

varunwho:
The filtering is not working in the example page, rest everything else is working. I'll make a PR, can you have a look?

[...]

PR #1481, opened 2021-09-21
Search bar filtering

varunwho:
Searchbar on the sidebar with filtering

Issue: #1446

### DIFF
--- a/website/src/components/SearchBar.js
+++ b/website/src/components/SearchBar.js
@@ -1,0 +1,12 @@
+import React from 'react'
+// import {useState} from 'react'
+
+function SearchBar(props) {
+    return (
+        <div>
+            <input className="text-black" type="text" placeholder="Search...." onChange={event => props.onChange(event.target.value)} />             
+        </div>
+    );
+};
+
+export default SearchBar
--- a/website/src/components/Sidebar.js
+++ b/website/src/components/Sidebar.js
@@ -1,10 +1,23 @@
-import React from 'react'
+import React, { useState } from 'react'
+import SearchBar from './SearchBar';
 
 function Sidebar ({children = []}) {
+    const [searchTerm,setSearchTerm] = useState(''); 
+    // console.log(children)
+    const filtered = 
+    children.filter(val => {
+        if (searchTerm === '') {
+            return val.props.text || val.props.children[0].props.text;
+        } else if (val.props.text.toLowerCase().includes(searchTerm.toLowerCase())) {
+            // console.log(val);
+            return val.props.text;
+        } 
+    })
     return (
         <div className="md:relative fixed animate__animated -left-full md:left-0 dark:bg-gray-700 bg-white shadow-md h-full md:block md:pt-0 md:w-4/12 top-0 w-9/12 z-20" id="sidebar">
             <div className="md:fixed h-full md:h-4/6 md:pt-0 md:w-4/12 overflow-y-auto pb-3 pl-5 pt-5">
-                {children}
+                <SearchBar onChange = {value => setSearchTerm(value) } />
+                {filtered}
             </div>
         </div>
     )
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -14605,9 +14605,9 @@
       }
     },
     "tailwindcss": {
-      "version": "npm:@tailwindcss/postcss7-compat@2.2.9",
-      "resolved": "https://registry.npmjs.org/@tailwindcss/postcss7-compat/-/postcss7-compat-2.2.9.tgz",
-      "integrity": "sha512-cE0ed0whqMPPih/mgo205e/Fdl/N9ZC8Z/0OyUBGWuvgNjNhsrT/avpuoYECI58FDTMT2tm5kKGJsJsl+sqh7A==",
+      "version": "npm:@tailwindcss/postcss7-compat@2.2.14",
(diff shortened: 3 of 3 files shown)


website/src/components/Sidebar.js
```diff
@@ -1,10 +1,23 @@
-import React from 'react'
+import React, { useState } from 'react'
+import SearchBar from './SearchBar';
 
 function Sidebar ({children = []}) {
+    const [searchTerm,setSearchTerm] = useState(''); 
+    // console.log(children)
+    const filtered = 
+    children.filter(val => {
```
shahednasser, line 8: This line throws an error when the website is being compiled. The reason is that you return a value inside the `if` and `else if` conditions, however, if an item does not comply with either of the issues nothing is returned.

So, at the end of the callback of `.filter` you should add a default return value.

website/src/components/SearchBar.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+// import {useState} from 'react'
+
+function SearchBar(props) {
```
shahednasser, line 4: Please change `props` to `{ onChange }` and update the use of `props` throughout the component accordingly.

website/src/components/Sidebar.js
```diff
@@ -1,10 +1,23 @@
-import React from 'react'
+import React, { useState } from 'react'
+import SearchBar from './SearchBar';
 
 function Sidebar ({children = []}) {
+    const [searchTerm,setSearchTerm] = useState(''); 
+    // console.log(children)
```
shahednasser, line 6: Remove this line

website/src/components/SearchBar.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+// import {useState} from 'react'
```
shahednasser, line 2: Remove this line

website/src/components/Sidebar.js
```diff
@@ -1,10 +1,23 @@
...
+        } else if (val.props.text.toLowerCase().includes(searchTerm.toLowerCase())) {
+            // console.log(val);
+            return val.props.text;
+        } 
+    })
     return (
         <div className="md:relative fixed animate__animated -left-full md:left-0 dark:bg-gray-700 bg-white shadow-md h-full md:block md:pt-0 md:w-4/12 top-0 w-9/12 z-20" id="sidebar">
             <div className="md:fixed h-full md:h-4/6 md:pt-0 md:w-4/12 overflow-y-auto pb-3 pl-5 pt-5">
-                {children}
+                <SearchBar onChange = {value => setSearchTerm(value) } />
```
shahednasser, line 19: Please remove the spaces around the = so it becomes like this `onChange={`

website/src/components/Sidebar.js
```diff
@@ -1,10 +1,23 @@
-import React from 'react'
+import React, { useState } from 'react'
+import SearchBar from './SearchBar';
 
 function Sidebar ({children = []}) {
+    const [searchTerm,setSearchTerm] = useState(''); 
+    // console.log(children)
+    const filtered = 
+    children.filter(val => {
+        if (searchTerm === '') {
```
shahednasser, line 9: I think if search term is empty then all values should be shown. So, you can return true right away without testing `val.props.text || val.props.children[0].props.text`. Unless there's a reason I'm not aware that it's necessary?

website/src/components/Sidebar.js
shahednasser: Also, this entire block `filtered = children.filtered` should be added inside `useEffect` which depends on `searchTerm`:

```js
useEffect(() => {
  //code here
}, [searchTerm]);
```

And I think `filtered` should be a state (and maybe renamed to searchResults):

```
const [searchResults, setSearchResults] = useState(children);
```

Then, inside `useEffect` that you will add, instead of:

```
const filtered = children.filter
```

You will use `setSearchResults` to set the new search results

shahednasser (changes requested):
Hi @varunwho thank you for your hard work! You did great, I just have a few comments to improve on your work.

Also, I wasn't able to actually see it in action as there was a compile error. You can see the notes below.

varunwho:
Thanks for letting me know @shahednasser, I'll try to resolve them (I'm a beginner in React though 😅)

shahednasser:
@varunwho don't worry you're doing a good job! You just need some fixes here and there, but you got the approach down right.

website/src/components/Sidebar.js
varunwho: > Also, this entire block `filtered = children.filtered` should be added inside `useEffect` which depends on `searchTerm`:
> 
> ```js
> useEffect(() => {
>   //code here
> }, [searchTerm]);
> ```
> 
> And I think `filtered` should be a state (and maybe renamed to searchResults):
> 
> ```
> const [searchResults, setSearchResults] = useState(children);
> ```
> 
> Then, inside `useEffect` that you will add, instead of:
> 
> ```
> const filtered = children.filter
> ```
> 
> You will use `setSearchResults` to set the new search results

I'll just learn useEffect and try to complete this in a day or two

varunwho:
@shahednasser I tried doing it but still it wasn't working, you can give the issue to someone else. Thankyou!

shahednasser:
Alright thank you for your work @varunwho

PR #1567, opened 2021-11-03
Adding Rotate Anticlockwise button

nisarga-kh:
Issue:
This PR is to add a new type of button called Rotate anticlockwise button. This button rotates in anticlockwise direction when hovered.


website/src/data/buttons.json
```diff
@@ -284,6 +284,13 @@
                 "type": "ripple",
                 "label": "Ripple",
                 "classes": "ripple-btn"
+            },
+            {
+                "type": "rotate-anticlockwise",
+                "label": "Rotate Anticlockwise",
+                "classes": "rotate-anticlockwise-btn" 
+
+
```
shahednasser, line 293: Remove these empty lines

shahednasser (changes requested):
Thank you @nisarga-kh but you need to make the following changes:

1. import `_rotate-anticlockwise.less` in `animated.less`
2. Run `npm run build`

After you do that please commit the changes so that we can see them on the website

nisarga-kh:
@shahednasser Have done the necessary modifications and have committed. Please do review once

shahednasser (changes requested):
Thank you @nisarga-kh for making the changes but the problem with this is that you're using in the animation the `top` and `left` attributes, but this button isn't absolute and it shouldn't be.

I suggest you use translate in the animation instead.

PR #1576, opened 2021-12-03
Sanket Formated some code

sanket95droid:
Changes: Formatted some code in dist. file.

Issue: Code format in dist.


shahednasser:
Thank you @sanket95droid for your work but formatting is not required for this file.

PR #1611, opened 2022-02-13
FIx # 1610 Issue

Rahul-go1503:
Issue: Redirect to ERROR page when click on license button in readme file


shahednasser:
Great thanks!

@all-contributors please add @Rahul-go1503 for doc

Issue #1609, opened 2022-02-12
Spotlight button's text color in dark mode


shahednasser:
In dark mode, [Spotlight's](https://sbuttons.netlify.app/examples#Spotlight) text color should be white before the animation as well.

JLambertazzo:
Hi! I'd like to work on this issue, can you please assign me?

Grandstaff79:
Hello, I would like to work on this issue as well, would you assign me ?

ej-sanmartin:
Since @crisboxing11 has successfully merged a change regarding this bug, issue should be closed! Did anyone check if this change, however, doesn't affect light mode?

ej-sanmartin:
Read the commits / merge history wrong, guess now would be a good time to test it out again

shahednasser:
This issue is still not resolved in case anyone is interested in working on it.

JLambertazzo:
I'll take it, can you assign me?

shahednasser:
Sure @JLambertazzo

JLambertazzo:
I can't get the changes to display on the example site.
Setting the button class to dark-mode or it's parent to have data-theme="dark" manually makes it work as expected but by default in dark mode it does not change.

shahednasser:
@JLambertazzo can you submit a PR so that I can see exactly what's going on?

JLambertazzo:
Created the pull request, let me know what you think

Poirei:
Can I join to work on this issue?

shahednasser:
@Poirei are you still interested in working on this?

lam-brian:
Hi, can I work on this issue? Also, quick question: Would you want white text on all the spotlight buttons except for the white ones? Otherwise when you hover it, the text would blend in with the background color.

shahednasser:
@lam-brian the text in dark mode should be white before the animation. As for after the animation you should just fix the white button to be another color

Dustpie:
Can I work on this issue please?

shahednasser:
Hello @Dustpie are you still interested in working on this?

Ruc08:
Hello @shahednasser, If no one is working on this issue then can I work on it?

shahednasser:
Hello @Ruc08 sure I'll assign you

Ruc08:
Thank you, @shahednasser! I'm starting to work on it now :)

Ruc08:
Hey @shahednasser, 

[...]

PR #1613, opened 2022-02-20
Spotlight Text Color Issue sButtons#1609

crisboxing11:
Fix: In dark mode, text color is now white before animation start.

Issue: Spotlight button's text color in dark mode #1609

### DIFF
--- a/src/components/animated/_spotlight.less
+++ b/src/components/animated/_spotlight.less
@@ -27,7 +27,7 @@
 }
 .sbtn.spotlight-btn {
   position: relative;
-  color: @darkText;
+  color: #fff;
   background-color: transparent;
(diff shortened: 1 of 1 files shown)


src/components/animated/_spotlight.less
```diff
@@ -27,7 +27,7 @@
 }
 .sbtn.spotlight-btn {
   position: relative;
-  color: @darkText;
+  color: #fff;
```
shahednasser, line 30: Please change to `@whiteText`

shahednasser (changes requested):
Hi @crisboxing11 thank you for your work!

Please take a look at my comment below. Also, make sure to run `npm run build` before you commit.

PR #1616, opened 2022-03-10
Multistate button

michalismichailidis1999:
I created a class for multistate buttons. For this to work button must have the class multi-state-btn and also
data-states="number of popup buttons". Then we can add any element except for buttons inside this button, and in order to make them popup when the button is clicked, they must have the state-btn class. Anything else inside the button will be its content (for example icon or text or both).


shahednasser (changes requested):
Thank you for your work @michalismichailidis1999 but you need to add the button in `website/src/data/buttons.json` to appear on the website. Please refer to the [contributing guidelines](https://github.com/sButtons/sbuttons/blob/master/CONTRIBUTING.md#instructions) to learn how to do that.

website/src/data/buttons.json
```diff
@@ -851,5 +851,23 @@
                 "usesColors": false
             }
         ]
+    },
+    {
```
shahednasser, line 855: You need to add a `children` key with an empty array if there are no sub types.

website/src/data/buttons.json
```diff
@@ -851,5 +851,23 @@
                 "usesColors": false
             }
         ]
+    },
+    {
+        "type": "multistate",
+        "label": "Multistate Buttons",
+        "classes": "multi-state-btn",
+        "data": [
```
shahednasser, line 859: What's this `data` key for? Please make sure you follow the structure detailed in the contributing guidelines.

shahednasser (changes requested):

website/src/data/buttons.json
michalismichailidis1999: The button will need the data-states="a number" attribute to work.

michalismichailidis1999: Should I remove the `data` key?

michalismichailidis1999 (commented):
I updated the buttons.json file and also changed the button so it doesn't need the data-state attribute

website/src/data/buttons.json
```diff
@@ -851,5 +851,63 @@
...
+        "children": [
+            {
+                "label": "State Button",
+                "type": "state-btn",
+                "classes": "state-btn",
+                "variations": [
+                    {
+                        "label": "2 States Button",
+                        "html": "{button context}<span>{button state content here}<span/><span>{button state content here}<span/>"
+                    },
```
shahednasser, line 892: What are these for? The html and label and all of that. I'm kind of confused on what they're for.

I suggest looking at the data for the basic buttons I think it might be similar to your use case.

shahednasser (changes requested):

website/src/data/buttons.json
michalismichailidis1999: the button should look like this

<button class="multi-state-btn states-2-btn">
     {an icon or text here}
     <span class="state-btn">{an icon or text here}</span>
     <span class="state-btn">{an icon or text here}</span>
<button/>

I didn't know exactly how to convert this into the buttons.json

PR #1634, opened 2022-07-05
Fixed spotlight button text color for dark mode

Ruc08:
This PR has fixed issue of spotlight buttons' text color for dark mode.

Issue: In dark mode, Spotlight buttons' text color should be white before the animation but it was dark.


shahednasser (changes requested):
Thank you @Ruc08 but you need to make the change to the button's styles in the `less` files, not make change to the React component that is rendering all components.

Ruc08:
Thank you @shahednasser for reviewing the PR! 

As of now I don't know less so give me few days so that I can learn it and make whatever changes are required.

